PR #101, opened 2016-08-26
Update Closure Compiler to 20160822.0.0

Dominator008:
Also update benchmark script to turn off most Closure checks and use
minimal externs. A JSDoc in react.js needs to be removed because it
currently causes an unsuppressible error in Closure.


hzoo:
cool thanks! is the .jar necessary to add?

PR #107, opened 2016-08-27
Capitalise ECMAScript and Babel

Daniel15:
Small tweaks to the readme: `ECMAScript` is a trademark and thus always needs to be capitalised that way when used. Babel is a proper noun and should be capitalised too.


hzoo:
Thanks!

PR #109, opened 2016-08-27
Remove old Mangle renamer

boopathi:


hzoo:
I don't particularly like leaving comments in the source - we should just remove it then?

kangax:
Agreed, it's all versioned so let's just remove.

boopathi:
The renamer is removed. The part that is turned to comments is the method "updateReferences" which a TODO for the current renamer.

Issue #112, opened 2016-08-27
minify-type-constructors shouldn’t rewrite `String(symbol)`


mathiasbynens:
## Input

``` js
var symbol = Symbol();
String(symbol);
// → 'Symbol()'
```
## Current output:

``` js
var symbol = Symbol();
symbol+"";
// → throws TypeError: Cannot convert a Symbol value to a string
```

mathiasbynens:
Btw, `String(x)` and `x + ""` (or `"" + x`) are not equivalent in general. Here’s another snippet that doesn’t use symbols but still produces different output after minification:

```js
var obj = {
  valueOf: _ => { console.log('valueOf'); return {}; },
  toString: _ => { console.log('toString'); return {}; }
};

try { String(obj); } catch (exception) { console.log(exception); }
// 1. toString
// 2. valueOf
// 3. TypeError: Cannot convert object to primitive value

// After minification, this becomes something like:
try { obj+""; } catch (exception) { console.log(exception); }
// 1. valueOf
// 2. toString
// 3. TypeError: Cannot convert object to primitive value
```

boopathi:
Thanks!

PR - #113 

Is it that this transformation shouldn't be applied to any pattern of `String(x)` ?

Since we don't know type `x` in `String(x)` during compile time, I guess we shouldn't transform this at all.

mathiasbynens:
I just wanted to make sure these edge cases are considered.

They might be rare enough in practice that the Babel team may decide not to care about them. But in that case, a note/warning in the README would be nice.

hzoo:
Yeah this is similar to template strings http://babeljs.io/docs/plugins/transform-es2015-template-literals/ although in this case users are intentionally using String.

boopathi:
This is merged and now there is an option to disable only `String()` type constructors.

PR #113, opened 2016-08-27
Don't transform String(x) to x + ''

boopathi:
#112


hzoo:
Should we just remove it or comment? I guess someone might think to add it again

boopathi:
I'm good for removing it. Thought it might be necessary to add some pattern that could be transformed later - and this would unbreak things that break

boopathi:
What about adding a comment about issue and removing the code ?

hzoo:
Cool 👍

kangax:
This should just be part of unsafe transform, which is what UglifyJS does. 99% of the time, it's used to convert number to string so we need to give people an option to use it so long as they're aware of consequences (which we should leave in notes). It also probably should be an opt-in, similar to GCC's advanced mode.

boopathi:
Hmm. Makes sense. I'm worried about people using Symbols - and I think it is used more and more now. Is there a way they can apply this transform without breaking symbols ? ... 

Like I want to apply the unsafe transformations (includes other type constructor transformations etc...) to most parts of the code except these places where I use `String(symbol)` - some hacky way at least ? and that I don't want to bail out completely from all the unsafe transformations because this one thing will break. I understand this argument stands for any unsafe transformation. But just want to know and document it if there is a way.

kangax:
Do we need options for plugins then? Could be something like `simplify-builtins: true | ["symbols", ...]`

boopathi:
Yeah. We could do something like - `includes` and `excludes` - 

with default `includes = ["bool", "array", "object", "number", "string"]` and `excludes = []`;
and excludes overrides includes, and users can specify what transformations they want to apply .. maybe ?

kangax:
Or make it a little friendlier and provide 3 options — all, exceptSymbols, none (as an example)

boopathi:
Oh... But we don't handle Symbol type at all ...

kangax:
@boopathi let's rebase this too? Also this isn't documented yet, right?

PR #118, opened 2016-08-28
Add link to repl [skip ci]

boopathi:


kangax:
Would be nice to be able to toggle separate plugins too.

boopathi:
We can enable passing in preset options once we get that ready ... maybe ?

kangax:
I mean in the UI it would be nice to be able to toggle them, similar to how you can toggle other presets right now.

hzoo:
Yeah would require some work. If we have that + umd plugins from npm we could debug more easily

PR #119, opened 2016-08-28
Fix bug - Conditionals transformation for impure paths

### LINKED ISSUE #115: Compression breaks commonmark npm package
I am using react-markdown which uses commonmark. When compiling the code with babili it breaks in some crazy way. It leads to an endless loop for both NodeJS and Chrome.

I am still investigating to narrow this down a little more.

Just wondering: Is there some special utf-8/unicode related functionality in babili?

boopathi:
+ (Close #115)


boopathi:
Will add a test for this .

kangax:
Looks good, thanks!

PR #127, opened 2016-08-29
Add error for babel-version <6.14.0

boopathi:
Okay. This time I'm sure it works.

But when we include preset options, I guess it is not possible to add them as well as support this.

/cc @loganfsmyth


boopathi:
I'm thinking of revisiting after we get somewhere with preset options #162 as when I tried previously, it was impossible to detect and throw this custom error (with a preset function). Also we don't have any complaints about this at all now.

kangax:
I'll close for now so that it doesn't distract us from other PR's

PR #128, opened 2016-08-29
add `--no-babelrc` option

hzoo:
Ref https://github.com/babel/babel.github.io/pull/898#discussion_r76549902 @loganfsmyth

### DIFF
--- a/packages/babili/src/index.js
+++ b/packages/babili/src/index.js
@@ -1,6 +1,6 @@
 import child from "child_process";
 
-const args = process.argv.slice(2).concat(["--presets=babili"]);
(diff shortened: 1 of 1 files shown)


boopathi:
But if I want to override this, is it possible ?

hzoo:
In the above case no. But then maybe we would say just use babel-cli then?

loganfsmyth:
I personally found it surprising that `babili` would pick up `.babelrc` configs. From a user standpoint, the fact that it's implemented with Babel seems like it should be an implementation detail. If you want to use a `.babelrc`, it seems to me like you should use the `babel` cli instead.

boopathi:
👍

Issue #76, opened 2016-08-02
Move some testing helpers into a module


hzoo:
Might be a little less setup?
Something like (maybe need to split up or move arguments around):
Basically always unpad, trim

``` js
const babel = require("babel-core");
const unpad = require("utils/unpad");

export function transform(source, plugin, options) {
  return babel.transform(unpad(source),  {
    plugins: [[plugin, options]],
  }).code.trim();
}

export function expectTransform(source, expected, plugin, options = {}) {
  return expect(transform(source, plugin, options)).toBe(unpad(expected));
}
```

boopathi:
Also, since we are always checking transformed input to static output, it would make sense to add `toMatchSnapshot` feature of jest, and we don't have to maintain a separate expected value at all.

I'm not sure if we can use it for any content - though in the [docs](https://facebook.github.io/jest/docs/api.html#tomatchsnapshot) it says React Component.

hzoo:
cc @cpojer?

cpojer:
Jest snapshots work with any serializable value, we just haven't documented it yet. This is actually the ideal use case for snapshots and should allow you guys to move much faster. Also cc @kyldvs who loves using it for strings.

boopathi:
Added in #606 . :)

PR #129, opened 2016-08-30
[WIP] Simplify Tests

boopathi:
#76

just wrote some codemod - and hand edited some edge cases. 
https://gist.github.com/boopathi/d979e5266df861052783acf9d3e61d22

Take a look before we apply to other packages


kangax:
My concern is that source and destination strings are in different places so it might be annoying to look back and forth.

@hzoo what do ya think

boopathi:
Yeah, I felt that too. So I thought we could do

```
expect({
  source,
  result: transform(source)
}).toMatchSnapshot()
```

hzoo:
A good point - being able to see the source/expected easily is useful

kangax:
Not sure if snapshot is the most appropriate approach here unless we can have input/output specified together.

hzoo:
Maybe we can leave this here and revisit.

Implementing what I wrote originally is easier and just reduces boilerplate

PR #132, opened 2016-08-30
Don't convert expression to expression in DCE

### LINKED ISSUE #130: Bad arrow function transform inside another function
As you can [see in the REPL](http://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2015%2Ces2015-loose%2Ces2016%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&experimental=false&loose=false&spec=false&code=function%20outer()%20%7B%0A%20%20const%20inner%20%3D%20(d)%20%3D%3E%20d.x%3B%0A%20%20return%20inner%3B%0A%7D%0A&playground=false), running babili on

```js
function outer() {
  const inner = (d) => d.x;
  return inner;
}
```
produces
```js
"use strict";function outer(){return function(a)a.x}
```

Which causes an `Unexpected identifier` error because of `function(a)a.x`.

```
const outer = () => {
  const inner = (d) => d.x;
  return inner;
}
```
also [fails in the same way](http://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2015%2Ces2015-loose%2Ces2016%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&experimental=false&loose=false&spec=false&code=const%20outer%20%3D%20()%20%3D%3E%20%7B%0A%20%20const%20inner%20%3D%20(d)%20%3D%3E%20d.x%3B%0A%20%20return%20inner%3B%0A%7D%0A&playground=false)

boopathi:
as it probably breaks arrowFunctions

+ (close #130)

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -1960,4 +1960,19 @@ describe("dce-plugin", () => {
     expect(transform(source)).toBe(expected);
   });
 
+  // https://github.com/babel/babili/issues/130
+  if("should not convert expression to expression during replace issue#130", () => {
+    const source = unpad(`
+      function outer() {
+        const inner = (d) => d.x;
+        return inner;
+      }
+    `);
+    const expected = unpad(`
+      function outer() {
+        return d => d.x;
+      }
+    `);
+    expect(transform(source)).toBe(expected);
+  });
 });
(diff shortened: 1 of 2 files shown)


boopathi:
Probably an upstream bug too - eg: http://astexplorer.net/#/3Fqc2CIwsu

boopathi:
Fix in babel - https://github.com/babel/babel/pull/3687

kangax:
Does it need Babel fix first?

boopathi:
No. It doesn't need babel fix. This reapplies the same fix as in Babel. So this can be merged :)

kangax:
👍

PR #134, opened 2016-08-31
Mangler performance - some improvement

boopathi:
This gives a ~=200ms of mangler perf. 

1. If NOT shouldConsiderSource, don't traverse Identifier & Literals for charset considerations
2. Don't recalculate hasBinding
3. loops

Most time was spent getting and verifying scope's bindings. 

The following difference is between `scope.hasBinding(next)` and `bindings.hop(next)`

```
run-mangle                 740.549  1        740.549       
scopable-loop              547.912  2222     0.247         
run-collect                370.834  1        370.834       
next-identifier            222.018  3693     0.060         
get-all-bindings           171.469  2222     0.077         
Collect-Identifier         65.521   19176    0.003         
compute-oldname            54.560   64879    0.001         
scopable-rename            49.586   3693     0.013         
rename                     42.556   3693     0.012         
```

and now after this 

```
run-mangle                 558.121  1        558.121       
scopable-loop              378.939  2222     0.171         
run-collect                369.423  1        369.423       
get-all-bindings           180.804  2222     0.081         
compute-oldname            53.859   64879    0.001         
scopable-rename            53.098   3693     0.014         
rename                     47.032   3693     0.013         
Collect-Identifier         44.661   19176    0.002         
next-identifier            39.646   3693     0.011         
```

### DIFF
--- a/packages/babel-plugin-minify-mangle-names/src/index.js
+++ b/packages/babel-plugin-minify-mangle-names/src/index.js
@@ -103,26 +103,37 @@ module.exports = ({ types: t }) => {
           //   i = 0;
           // }
 
-          Object
-            .keys(scope.getAllBindings())
-            .filter((b) => {
-              const binding = scope.getBinding(b);
+          const bindings = scope.getAllBindings();
+          const names = Object.keys(bindings);
 
-              return scope.hasOwnBinding(b)
-                && !binding.path.isLabeledStatement()
-                && !mangler.isBlacklist(b, mangler.blacklist)
-                && (mangler.keepFnames ? !isFunction(binding.path) : true);
-            })
-            .map((b) => {
-              let next;
-              do {
-                next = getNext();
-              } while (!t.isValidIdentifier(next) || scope.hasBinding(next) || scope.hasGlobal(next) || scope.hasReference(next));
-              // TODO:
-              // re-enable this
-              // resetNext();
-              mangler.rename(scope, b, next);
-            });
+          for (let i = 0; i < names.length; i++) {
+            const oldName = names[i];
+            const binding = bindings[oldName];
+
+            if (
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-mangle-names/src/index.js
```diff
@@ -103,26 +107,37 @@ module.exports = ({ types: t }) => {
...
+              continue;
+            }
+
+            let next;
+            do {
+              next = getNext();
+            } while (
+              !t.isValidIdentifier(next)
+              || hop.call(bindings, next)
+              || scope.hasGlobal(next)
```
kangax: Idea: collect global references upfront and check manually instead of calling `hasGlobal` for each

boopathi: each scope maintains its global references when the scope is created only (I think), and we visit scopes only once - and we transform in place. So my gut feel is that it's not possible to do better that that.

kangax: Wait, all scopes end up in the global... why do they need to store references?

boopathi: I'm not sure. `scope is created (globals calculated from bindings) -> visitor applied` is what I thought. I'll check though.

boopathi: Also, from the timing values, the `next-identifier` is just 39ms - this includes -

```js
let next;
            do {
              next = getNext();
            } while (
              !t.isValidIdentifier(next)
              || hop.call(bindings, next)
              || scope.hasGlobal(next)
              || scope.hasReference(next)
            );
```

So this is not something we need to worry about ...

PR #136, opened 2016-08-31
add changelog [skip ci]

hzoo:
Generated from https://github.com/lerna/lerna-changelog

Probably want to modify it though, since it doesn't make sense with independent packages


hzoo:
We can keep it for now and fix it up later

PR #146, opened 2016-09-03
enhansed live example link

tomByrer:
Before, the `[] Minify (Babili)` checkbox wasn't checked.  Now it is, & includes the example code below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 - babili is consumable via API, CLI, or babel preset.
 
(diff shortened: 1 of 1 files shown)


tomByrer:
I suggest to wait for https://github.com/babel/babel.github.io/issues/917 to be fixed first.

hzoo:
I think you were getting the confusion from before because I didn't want to add example code but I guess we should.

tomByrer:
Thanks, yes I needed a more 'idiot proof' example :)

hzoo:
kk makes sense, I'm sure it was confusing to others as well

Issue #54, opened 2016-07-19
Minify options


boopathi:
Since it is a preset, there won't be any options passed to the minifier. 

for example,
- mangle topLevel names
- mangle eval accessible bindings - don't deopt because of the presense of eval
- apply unsafe optimizations like - `!==` to `!=`, `>=` to `<`
- drop debugger & console statements or just one of them
- preserve fn args and fn name for code depending on fn.length & fn.name

etc...

How to support these as options to be passed to the minifier - when used as a CLI or a Preset?

jacott:
I would much appreciate a `keepClassNames` option

PR #148, opened 2016-09-05
Improve benchmarks accuracy

kangax:
This changes benchmark suite to run tests multiple times and get average. It also introduces "best size" and "best speed" versions of babili.


package.json
```diff
@@ -47,6 +47,22 @@
...
+    "babel-plugin-minify-mangle-names": "^0.0.2",
+    "babel-plugin-minify-replace": "^0.0.1",
+    "babel-plugin-minify-simplify": "^0.0.2",
+    "babel-plugin-minify-type-constructors": "^0.0.1",
+    "babel-plugin-transform-member-expression-literals": "^6.8.0",
+    "babel-plugin-transform-merge-sibling-variables": "^6.8.0",
+    "babel-plugin-transform-minify-booleans": "^6.8.0",
+    "babel-plugin-transform-property-literals": "^6.8.0",
+    "babel-plugin-transform-simplify-comparison-operators": "^6.8.0",
+    "babel-plugin-transform-undefined-to-void": "^6.8.0"
```
kangax: I'm not sure if we can avoid this? Needed for benchmarks to be able to modify plugin options (such as simplify).

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -333,7 +333,12 @@ module.exports = ({ types: t }) => {
       Function: {
         enter: earlyReturnTransform,
 
-        exit(path) {
+        exit(path, state) {
+          const multiPass = state.opts.multiPass === false ? false : true;
```
kangax: This is true by default which probably makes sense most of the time.

scripts/benchmark.js
```diff
@@ -64,24 +67,82 @@ function test(name, callback) {
...
+      ],
+      require("babel-plugin-minify-type-constructors"),
+      require("babel-plugin-transform-member-expression-literals"),
+      require("babel-plugin-transform-merge-sibling-variables"),
+      require("babel-plugin-transform-minify-booleans"),
+      require("babel-plugin-transform-property-literals"),
+      require("babel-plugin-transform-simplify-comparison-operators"),
+      require("babel-plugin-transform-undefined-to-void"),
+    ],
+    comments: false,
```
kangax: Not too happy with this duplication but as options grow, it will diverge more.

Or should we create 2 proper separate presets?

packages/babel-plugin-minify-simplify/src/index.js
boopathi:
```js
exit(path, { opts: { multipass = true } = {} } = {}) {
}
```

?

scripts/benchmark.js
boopathi: we should enable preset options #54

kangax: Ok, let's wait for that to merge then.

packages/babel-plugin-minify-simplify/src/index.js
kangax: A bit too cryptic for my taste :) What do you think?

boopathi: For this (putting everything in one line and just one 2-level-deep option) it looks cryptic, actually using this is much clearer sometimes, We could use default destructuring assignment when we have more options.

scripts/benchmark.js
boopathi: Or we can land this as two separate things. Multipass option in the plugin and benchmarks in another.

kangax:
@boopathi now that preset options are in, I guess let me change PR to use that

scripts/benchmark.js
```diff
@@ -116,23 +119,54 @@ function test(name, callback) {
...
+    }).code;
+  });
+
+  test("babili (best size)", function (code) {
+    return babel.transform(code, {
+      sourceType: "script",
+      presets: [
+        [require("../packages/babel-preset-babili"), {
+          simplify: {
+            multiPass: true
```
kangax: @boopathi this doesn't seem to show any difference in any of the files I'm testing...

boopathi: there is no multipass option implemented in simplify.

boopathi: You implemented it here - in this commit - https://github.com/babel/babili/pull/148/commits/827a8356545e285cd3cbafeb85ea44e909e135d5#diff-05684bffcab704dd5d3842adbf921357 and your latest changes don't seem to have this one

kangax: Right, I'm an idiot, was thinking that we already added it.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -507,7 +507,12 @@ module.exports = ({ types: t }) => {
...
+          if (!multiPass) {
+            return;
+          }
+
           earlyReturnTransform(path);
 
           if (!path.node[shouldRevisit]) {
```
boopathi: Doesn't `multipass` affect this ?

It throws an error for me - `Container is falsy`. I suppose this happens when calling `path.replaceWith` on the path whose node is modified.

boopathi: I tried the script from the readme - 

```
./scripts/benchmark.js react@0.14.3 react/dist/react.js
```

boopathi:
Also, the runtime metric looks inverted ?

```
                    raw     raw win gzip    gzip win parse time run time (average)
babili (best size)  34.36kB 262%    12.03kB 147%     2ms        698ms
babili (best speed) 34.37kB 262%    12.03kB 147%     2ms        947ms
```

kangax:
@boopathi I just remembered that you removed `Function: { enter: ... }` (https://github.com/babel/babili/blob/4c4c10f42c8071c8b9ade0b92bb31b566008e204/packages/babel-plugin-minify-simplify/src/index.js) so that's probably why there's no difference now?

boopathi:
Ah. yeah. #282

kangax:
How about I remove multiPass for now, and merge the rest? That way we can pass options through presets in a benchmark since we'll likely have them in a future.

boopathi:
Cool! This means we run each benchmark 3 times .. right?

kangax:
Yep. We can also expose it via command line flag `./scripts/benchmark.js ... --runs=10`

PR #155, opened 2016-09-11
Fix DCE for ArrayPatterns and ObjectPatterns

### LINKED ISSUE #151: Incorrect behaviour with array destructuring
**Babili** seems to be misinterpreting array destructuring inside function bodies:

### Input
```js
const me = lyfe => {
  const [swag] = lyfe;
  return swag;
}
```

### Output
```js
const me=a=>{return a};
```

[REPL](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2016&code=const%20me%20%3D%20lyfe%20%3D%3E%20%7B%0D%0A%20%20const%20%5Bswag%5D%20%3D%20lyfe%3B%0D%0A%0D%0A%20%20return%20swag%3B%0D%0A%7D)

As you can see, it acts as if there's no destructuring going on at all.

--

It fails if you try to get more than one item:

### Input

```js
const me = lyfe => {
  const [swag, yolo] = lyfe;

  return swag && yolo;
}
```

### Error message

```
repl: NodePath has been removed so is read-only. (This is an error on an internal node. Probably an internal error)
```

[REPL](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2016&code=const%20me%20%3D%20lyfe%20%3D%3E%20%7B%0D%0A%20%20const%20%5Bswag%2C%20yolo%5D%20%3D%20lyfe%3B%0D%0A%0D%0A%20%20return%20swag%20%26%26%20yolo%3B%0D%0A%7D)

--

On the other hand, destructuring outside functions seems to be fine. Object destructuring also worked for me.

boopathi:
+ Bail out for replacement of single use variable inside a function

TODO (for later): maybe a more intelligent approach to replace - for ex,

```js
// from
const [a] = b; return a;
const {a} = b; return a;

// to
return b[0];
return b.a;
```
+ (Close #151)

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -1975,4 +1975,28 @@ describe("dce-plugin", () => {
     `);
     expect(transform(source)).toBe(expected);
   });
+
+  // https://github.com/babel/babili/issues/151
+  it("should fix issue#151 - array patterns and object patterns", () => {
+    const source = unpad(`
+      const me = lyfe => {
+        const [swag] = lyfe;
+        return swag;
+      };
+    `);
+    const expected = source;
+    expect(transform(source)).toBe(expected);
+  });
+
+  // https://github.com/babel/babili/issues/151
+  it("should fix issue#151 - array patterns and object patterns 2", () => {
+    const source = unpad(`
+      const me = lyfe => {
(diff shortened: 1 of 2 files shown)


loganfsmyth:
`const [a] = b; return a;` isn't the same as `b[0];`, that'll break for Map and Set and such.

boopathi:
Yeah! Forgot about those. Thanks.

boopathi:
I think this PR is good to go to fix the bug. Will investigate later what other transformations can be applied.

PR #157, opened 2016-09-12
don't merge block-scoped sibling vars across scopes, fix #153

### LINKED ISSUE #153: Improper let scoping
```JavaScript
let x = 0;

for(let i = 0; i < 10; ++i) {
	x++;
}

console.log(x);
```
is converted to:
```JavaScript
for(let x=0,a=0;10>a;++a)x++;console.log(x);
```

**x** is being moved into the first part of the **for** loop which causes it to be scoped only to the **for block**. The call to **console.log(x)** is outside of the block and thus throws a **ReferrenceError**.

goto-bus-stop:
Only move `var` declarations into `for` loop initialisers, so `let` and `const`
declarations are still available outside the loop.

### DIFF
--- a/packages/babel-plugin-transform-merge-sibling-variables/__tests__/transform-merge-sibling-variables-test.js
+++ b/packages/babel-plugin-transform-merge-sibling-variables/__tests__/transform-merge-sibling-variables-test.js
@@ -36,4 +36,38 @@ describe("transform-merge-sibling-variables-plugin", () => {
 
     expect(transform(source).trim()).toBe(expected);
   });
+
+  it("don't concat block-scoped variables in for loops", () => {
+    const source = unpad(`
+      let i = 0;
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+
+    expect(transform(source)).toBe(source);
+  });
+
+  it("don't concat constants in for loops", () => {
+    const source = unpad(`
+      const j = 0;
+      for (const x = 0;;) console.log(j + x);
+    `);
+
+    expect(transform(source)).toBe(source);
+  });
+
+  it("concat block-scoped vars next to, but not into for loops", () => {
+    const source = unpad(`
+      let i = 0;
+      let y = 0;
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+    const expected = unpad(`
+      let i = 0,
+          y = 0;
+
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+
+    expect(transform(source)).toBe(expected);
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-transform-merge-sibling-variables/__tests__/transform-merge-sibling-variables-test.js
```diff
@@ -36,4 +36,33 @@ describe("transform-merge-sibling-variables-plugin", () => {
...
+    const source = unpad(`
+      let i = 0;
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+    const expected = unpad(`
+      let i = 0;
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+
+    expect(transform(source).trim()).toBe(expected);
```
hzoo: Do we just reuse the source if it's the same? Forgot what other tests have done (in this case you could `expect(transform(source).trim()).toBe(source);`

goto-bus-stop: Ah! yes, other tests reuse the `source`. fixing, thanks!

hzoo:
I don't think you could put a const in a for loop anyway since it's assigning in a loop, we could add a test for coverage though (for `const`)

goto-bus-stop:
Sounds good! added. ty for review! :smile:

packages/babel-plugin-transform-merge-sibling-variables/__tests__/transform-merge-sibling-variables-test.js
```diff
@@ -36,4 +36,38 @@ describe("transform-merge-sibling-variables-plugin", () => {
...
     expect(transform(source).trim()).toBe(expected);
   });
+
+  it("don't concat block-scoped variables in for loops", () => {
+    const source = unpad(`
+      let i = 0;
+      for (let x = 0; x < 10; x++) console.log(i + x);
+    `);
+
+    expect(transform(source).trim()).toBe(source);
```
boopathi: The trim is not necessary here. should work without it.

goto-bus-stop: Removed, thanks!

kangax:
This broke tests?

hzoo:
It did? looks like it was passing before at least on travis?

hzoo:
It was an issue with checkout?

$ git checkout -qf 24527c774ed7777046930c5c4e8194c6caed2eb1
fatal: reference is not a tree: 24527c774ed7777046930c5c4e8194c6caed2eb1
The command "git checkout -qf 24527c774ed7777046930c5c4e8194c6caed2eb1" failed and exited with 128 during .

Issue #159, opened 2016-09-13
Breaks Magnific Popup gallery


callym:
Using Babili with Magnific Popup breaks the **gallery** view, but not the rest of the script.

Using the script from [here](http://dimsemenov.com/plugins/magnific-popup/)
https://codepen.io/callym/pen/JRGawr

Replacing that script with the minified output from babili
https://codepen.io/callym/pen/jrWvJW

hzoo:
Thanks for the report, but we'll need to have a reduced reproduction step. Right now we just know that something in that whole file isn't working. And did you just paste it into the repl?

callym:
I've been trying to replace sections of the unminified file with the minified one, and the function that causes it is:
```javascript
		initGallery: function() {

			var gSt = mfp.st.gallery,
				ns = '.mfp-gallery';

			mfp.direction = true; // true - next, false - prev

			if(!gSt || !gSt.enabled ) return false;

			_wrapClasses += ' mfp-gallery';

			_mfpOn(OPEN_EVENT+ns, function() {

				if(gSt.navigateByImgClick) {
					mfp.wrap.on('click'+ns, '.mfp-img', function() {
						if(mfp.items.length > 1) {
							mfp.next();
							return false;
						}
					});
				}

				_document.on('keydown'+ns, function(e) {
					if (e.keyCode === 37) {
						mfp.prev();
					} else if (e.keyCode === 39) {
						mfp.next();
					}
				});
			});

			_mfpOn('UpdateStatus'+ns, function(e, data) {
				if(data.text) {
					data.text = _replaceCurrTotal(data.text, mfp.currItem.index, mfp.items.length);
				}
			});

			_mfpOn(MARKUP_PARSE_EVENT+ns, function(e, element, values, item) {
				var l = mfp.items.length;
				values.counter = l > 1 ? _replaceCurrTotal(gSt.tCounter, item.index, l) : '';
			});

			_mfpOn('BuildControls' + ns, function() {
				if(mfp.items.length > 1 && gSt.arrows && !mfp.arrowLeft) {
					var markup = gSt.arrowMarkup,
						arrowLeft = mfp.arrowLeft = $( markup.replace(/%title%/gi, gSt.tPrev).replace(/%dir%/gi, 'left') ).addClass(PREVENT_CLOSE_CLASS),
[...]

PR #160, opened 2016-09-14
don't remove impure reachable parts of logical expressions

goto-bus-stop:
An attempt at dealing with #159.

Impure statements like function calls were removed from the right-hand side of `&&` expressions if they were inferred to return a falsy value (e.g. when using the `void` operator).

This patch checks if the thing to be removed is pure. There's a lot more manual checking involved but it also simplifies more stuff.

### DIFF
--- a/packages/babel-plugin-minify-guarded-expressions/__tests__/guarded-expressions-test.js
+++ b/packages/babel-plugin-minify-guarded-expressions/__tests__/guarded-expressions-test.js
@@ -71,11 +71,58 @@ describe("guarded-expressions-plugin", () => {
       alert("");
     `);
     expect(transform(source)).toBe(expected);
+
+    source = unpad(`
+      alert(new Foo() || false);
+    `);
+    expected = unpad(`
+      alert(new Foo());
+    `);
+    expect(transform(source)).toBe(expected);
   });
 
-  it("should not remove unpure statements that always evaluate to false", () => {
-    const source = unpad(`
+  it("should simplify truthy expressions", () => {
+    let source = unpad(`
+      alert(1 && new Bar());
+    `);
+    let expected = unpad(`
+      alert(new Bar());
+    `);
+    expect(transform(source)).toBe(expected);
+
+    source = unpad(`
+      alert(true && new Bar());
+    `);
+    expected = unpad(`
+      alert(new Bar());
+    `);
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-guarded-expressions/src/index.js
```diff
@@ -15,8 +15,17 @@ module.exports = function({ types: t }) {
...
-              path.replaceWith(node.left);
+            const left = path.get("left");
+            const right = path.get("right");
+            if (node.operator === "&&") {
+              if ((right.evaluateTruthy() === false && right.isPure()) ||
+                  (left.evaluateTruthy() === false && left.isPure())) {
+                path.replaceWith(node.left);
+              }
+            } else if (node.operator === "||") {
+              if (left.evaluateTruthy() === false && left.isPure()) {
```
goto-bus-stop: whoops, that's not right :P

goto-bus-stop:
I think I fixed my logic mistakes. I'm not entirely sure if all the branches I put in are necessary or if `flip-expressions` obsoletes some of them.

packages/babel-plugin-minify-guarded-expressions/__tests__/guarded-expressions-test.js
```diff
@@ -71,5 +71,59 @@ describe("guarded-expressions-plugin", () => {
...
+  });
+
+  it("should not remove reachable impure statements", () => {
+    let source = unpad(`
+      a && void alert('Side effect');
+    `);
+    expect(transform(source)).toBe(source);
+
+    source = unpad(`
+      alert(func() || true);
```
kangax: Shouldn't we remove `true` from the RHS here since it has no side effects?

boopathi: `a = foo() || true`. `a` should be `true` if `foo()` is truthy.

goto-bus-stop: if `func()` is falsy, `func() || true` would still evaluate to `true`, so the RHS is significant for the final value

kangax: Oh, of course. Missed it.

PR #169, opened 2016-09-29
[DCE] Fix var hoisting

### LINKED ISSUE #167: var hoisting
### Input:

```js
function foo() {
  a = 1;
  return a;
  var a;
}
```

### Actual Output:
No var decl.

```js
function foo(){return b=1,b}
```

[REPL](http://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=es2017&code=function%20foo()%20%7B%0A%20%20a%20%3D%201%3B%0A%20%20return%20a%3B%0A%20%20var%20a%3B%0A%7D&experimental=false&loose=false&spec=false&playground=true)

### Expected:
vars should be hoisted

and the output should become `function foo(){return 1}`

boopathi:
+ (Close #167)

One other option to handle this will be to hoist var declarations as one of the first plugins. But @amasad had concerns with the hoisting effect on gzip in #52 . So this approach is okay IMO. 

`a = 1; return a; var a;` could not be reduced to `return 1` because we don't have the flow of control in the program - this is here in babel - https://github.com/babel/babel/issues/4420

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -283,7 +283,7 @@ module.exports = ({ types: t, traverse }) => {
           continue;
         }
 
-        if (purge && !p.isFunctionDeclaration()) {
+        if (purge && !canExistAfterCompletion(p)) {
           removeOrVoid(p);
         }
       }
@@ -299,7 +299,7 @@ module.exports = ({ types: t, traverse }) => {
 
       // Not last in it's block? (See BlockStatement visitor)
       if (path.container.length - 1 !== path.key &&
-          !path.getSibling(path.key + 1).isFunctionDeclaration() &&
+          !canExistAfterCompletion(path.getSibling(path.key + 1)) &&
           path.parentPath.isBlockStatement()
       ) {
         // This is probably a new oppurtinity by some other transform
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -895,4 +895,10 @@ module.exports = ({ types: t, traverse }) => {
       };
     }
   }
+
+  // things that are hoisted
+  function canExistAfterCompletion(path) {
```
kangax: So basically anything that doesn't conform to TDZ? What about function?

```js
a = 1;
return a;
function a() { }
```

boopathi: That test is already covered in `should be fine with fun decl after return` if that's what you meant.

boopathi: https://github.com/babel/babili/blob/32c809b39638aba7d2bcb7bec684f91f2f551be3/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js#L490

kangax: Ah, perfect!

PR #179, opened 2016-10-05
Fix guarded expressions

### LINKED ISSUE #171: minify-simplify + minify-guarded-expressions incorrectly transform a switch statement
I've distilled my repro case to this:
```js
function foo(bar) {
	bar = bar || 'baz';

	switch (bar) {
		case 'spam':
			return true;
		case 'ham':
		case 'eggs':
			return false;
	}
	return false;
}
```
Notice that the function only returns `true` if `bar == 'spam'`.

When I run it through `minify-simplify` plugin (`babel --plugins=minify-simplify`) it produces a correct result:
```js
function foo(bar) {
	return bar = bar || 'baz', bar === 'spam' || (bar === 'ham' || bar === 'eggs') && false;
}
```

When I run the source through `minify-guarded-expressions` separately (`babel --plugins=minify-guarded-expressions`) it has no effect.

However when I run `minify-simplify` and `minify-guarded-expressions` at the same time (`babel --plugins=minify-guarded-expressions,minify-simplify`) I get this:
```js
function foo(bar) {
	return bar = bar || 'baz', bar === 'spam' || bar === 'ham' || bar === 'eggs';
}
```
Notice that the function now returns `true` for all the cases, not just `'spam'`.

Here's a [REPL](http://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=true&presets=es2015&code=function%20foo(bar)%20%7B%0A%20%20bar%20%3D%20bar%20%7C%7C%20%27baz%27%3B%0A%0A%20%20switch%20(bar)%20%7B%0A%20%20%20%20case%20%27spam%27%3A%0A%20%20%20%20%20%20return%20true%3B%0A%20%20%20%20case%20%27ham%27%3A%0A%20%20%20%20case%20%27eggs%27%3A%0A%20%20%20%20%20%20return%20false%3B%0A%20%20%7D%0A%20%20return%20false%3B%0A%7D)

### LINKED ISSUE #174: minify-guarded-expressions is too agressive
Input:
```js
foo = (bar.baz || '').length;
```
Output with `minify-guarded-expressions` enabled:
```js
foo = bar.baz.length;
```
The problem is that `bar.baz` can be undefined, so `bar.baz.length` will throw an error.

For some reason I can't reproduce it on the [REPL page](http://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=true&presets=latest&code=foo%20%3D%20(bar.baz%20%7C%7C%20%27%27).length%3B), but it is reproducible via CLI: `babel --plugins=minify-guarded-expressions`

### LINKED ISSUE #176: babili overminifies statements with || ''
With version 0.0.4 of babel-preset-babili, Safe setting a variable to a null string is removed by the minification process. 

The following code:

```
this.value = a || '';
```

is minified to:

```
this.value = a;
```

boopathi:
+ (close #171)
+ (close #174)
+ (close #176)

### DIFF
--- a/packages/babel-plugin-minify-guarded-expressions/src/index.js
+++ b/packages/babel-plugin-minify-guarded-expressions/src/index.js
@@ -17,6 +17,16 @@ module.exports = function({ types: t }) {
 
             const left = path.get("left");
             const right = path.get("right");
+
+            // issues - 171, 174, 176
+            // we assume that it is returned/assigned/part of a bigger expression
+            // or utilized somehow
+            // we check if we shouldBail only when evaluating
+            // the rightside of the expression;
+            // if the left side is evaluated to be deterministic,
+            // we can safely replace the entire expression
+            const shouldBail = !path.parentPath.isExpressionStatement();
+
             if (node.operator === "&&") {
               const leftTruthy = left.evaluateTruthy();
               if (leftTruthy === false) {
@@ -24,7 +34,7 @@ module.exports = function({ types: t }) {
                 path.replaceWith(node.left);
               } else if (leftTruthy === true && left.isPure()) {
                 path.replaceWith(node.right);
-              } else if (right.evaluateTruthy() === false && right.isPure()) {
+              } else if (right.evaluateTruthy() === false && right.isPure() && !shouldBail) {
                 path.replaceWith(node.left);
               }
             } else if (node.operator === "||") {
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-guarded-expressions/src/index.js
```diff
@@ -17,14 +17,24 @@ module.exports = function({ types: t }) {
...
             const right = path.get("right");
+
+            // issues - 171, 174, 176
+            // we assume that it is returned/assigned/part of a bigger expression
+            // or utilized somehow
+            // we check if we shouldBail only when evaluating
+            // the rightside of the expression;
+            // if the left side is evaluated to be deterministic,
+            // we can safely replace the entire expression
+            const shouldBail = !path.parentPath.isExpressionStatement();
```
boopathi: I hope this assumption doesn't have any corner cases that fail. Will add test cases once this seems to work out well for all corner cases.

kangax:
Looks reasonable, thanks!

PR #182, opened 2016-10-08
Remove unused fn params

### LINKED ISSUE #163: babel-plugin-minify-dead-code-elimination should remove unused arguments
Original issue: https://github.com/babel/babel/issues/4531

### Input Code

Here's [a reproduction repo](https://github.com/kentcdodds/temp-babel-issue-dce)

```js
const foo = (arg1, arg2) => arg1
```

**.babelrc** (or what it might look like)

```js
{
  "plugins": ["minify-dead-code-elimination"]
}
```

### Expected Behavior

I would expect the output to be:

```js
const foo = (arg1) => arg1
```

### Current Behavior

Instead, the plugin doesn't do anything with unused arguments

```js
const foo = (arg1, arg2) => arg1
```

### Possible Solution

I'm not certain whether this is totally the responsibility of `babel-plugin-minify-dead-code-elimination`, but I can't think of any other plugin that should do this.

### Context

I'm writing a program slicing tool that's using `babel-plugin-minify-dead-code-elimination` directly to remove dead code after I've manipulated the AST to remove sliced nodes.

### Your Environment

| software         | version
| ---------------- | -------
| Babel            |  6.14.0
| node             | 6.5.0
| npm              | 3.10.7
| Operating System | Mac OSX

boopathi:
+ (close #163)
+ Introduces a new option for DCE - `keepFnArgs`
+ Renames - `keepFnames` to `keepFnName` for DCE

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -121,12 +121,7 @@ module.exports = ({ types: t, traverse }) => {
               break;
             }
 
-            Object.defineProperty(binding, markForRemoval, {
-              enumerable: false,
-              value: true,
-              writable: false,
-              configurable: true
-            });
+            binding[markForRemoval] = true;
             continue;
           } else if (param.isAssignmentPattern()) {
             const left = param.get("left");
@@ -140,12 +135,7 @@ module.exports = ({ types: t, traverse }) => {
                 break;
               }
 
-              Object.defineProperty(binding, markForRemoval, {
-                enumerable: false,
-                value: true,
-                writable: false,
-                configurable: true
-              });
+              binding[markForRemoval] = true;
               continue;
             }
           }
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -6,6 +6,11 @@ module.exports = ({ types: t, traverse }) => {
   const removeOrVoid = require("babel-helper-remove-or-void")(t);
   const shouldRevisit = Symbol("shouldRevisit");
 
+  // this is used for tracking fn params that can be removed
+  // as traversal takes place from left and
+  // unused params can be removed only on the right
+  const markForRemoval = Symbol("markForRemoval");
+
   const main = {
     // remove side effectless statement
     ExpressionStatement(path) {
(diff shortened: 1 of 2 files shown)


boopathi:
The tests fail right now because I've set the default as to remove unused fn params and haven't modified other tests. Should it be opt-in instead of opt-out ?

kangax:
I think `keepFnArgs` would be better and then we should rename `keepFname` to `keepFnName`

kangax:
Removing them by default seems fine if we're doing same with names. IMO, if we start getting complains we can change.

boopathi:
Cool - `keepFnArgs` and `keepFnName`.

Hit a bug - `function foo(a,b) { b }` -> `function foo(b) { b }` which is totally wrong.

boopathi:
Updated.

kangax:
Great. Could you update benchmarks table as well.

boopathi:
The diff is 

```diff
-uglify     21.79kB 221%    7.29kB 169%     1ms        247ms
-closure    21.67kB 223%    7.37kB 167%     2ms        1223ms
-babili     21.9kB  219%    7.46kB 164%     2ms        722ms
-closure js 24.01kB 191%    8.04kB 144%     2ms        3301ms
+uglify     21.79kB 221%    7.29kB 169%     2ms        387ms
+closure    21.67kB 223%    7.37kB 167%     2ms        3195ms
+babili     21.86kB 220%    7.46kB 163%     2ms        943ms
+closure js 24.01kB 191%    8.04kB 144%     2ms        4943ms
```

we need to move running benchmarks on a machine with same config. maybe on CI?

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -6,6 +6,11 @@ module.exports = ({ types: t, traverse }) => {
   const removeOrVoid = require("babel-helper-remove-or-void")(t);
   const shouldRevisit = Symbol("shouldRevisit");
 
+  // this is used for tracking fn params that can be removed
+  // as traversal takes place from left and
+  // unused params can be removed only on the right
+  const markForRemoval = Symbol("markForRemoval");
```
kangax: Nit: shouldn't it be `markedForRemoval`?

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -100,17 +105,71 @@ module.exports = ({ types: t, traverse }) => {
...
+
+          if (param.isIdentifier()) {
+            const binding = scope.bindings[param.node.name];
+            if (binding.referenced) {
+              // when the first binding is referenced (right to left)
+              // exit without marking anything after this
+              break;
+            }
+
+            Object.defineProperty(binding, markForRemoval, {
```
kangax: Do we need to `defineProperty` here instead of just assigning it? Is this because we don't want to taint binding with "irrelevant" enumerable prop? I'm just worried about performance.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
boopathi: hmmm. IMO, depends on context, 

at the place of assignment, mark this for removal,
at the place of verifying, is this marked for removal,

Not sure. IMO, either way should be fine.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
boopathi: Ohhh . defineProp has perf drawbacks!! didn't know. I just didn't want it to be enumerable.

hzoo:
We should update the readme when possible when changing options + explanation

PR #184, opened 2016-10-11
minify-dead-code-elimination: remove dead logical expressions

christophehurpeau:
Transforms:

```js
false && foo();
true && bar();
```

into:

```js
false;
bar();
```


christophehurpeau:
@boopathi do you have something in mind ? How big ?

boopathi:
+ Logical expr as a part of other expr like fn call,
+ expr with falsy & truthy values (not just true and false) - like '' || foo ...

boopathi:
How different is this from this code in minify-guarded-expressions - https://github.com/babel/babili/blob/master/packages/babel-plugin-minify-guarded-expressions/src/index.js#L14-L51 Doesn't minify-guarded-expressions already take care of this optimization.

christophehurpeau:
@boopathi okay thanks ! Did not saw it.

PR #186, opened 2016-10-13
Mangle - Fix label shadowing

### LINKED ISSUE #185: Error mangling identifier in try-catch when there's same-named label
This:

```js
function f(a) {
  try {
    a: {
      console.log(a);
    }
  } catch ($a) { }
}
```

becomes:

```js
"use strict";function f(b){try{a:console.log(a)}catch(c){}}
```

Notice `console.log(a)` instead of `console.log(b)`.

boopathi:
+ (Close #185) 

Not sure if there are any other bugs that this will lead to. If so, afaik, it will only lead to LabelStatement bugs.

+ Adds a 250-300ms to mangler because of the double pass. Maybe we can add a flag to `useLabels` - maybe ? and when babel starts treating Labels in a separate namespace, we can remove the code - incase we merge this...
+ this needs traversing program twice because the second time, the vars get replaced from outer scope and we cannot bind the referencedIdentifier to its BindingIdentifier - as we check this based on the Identifier name

### DIFF
--- a/packages/babel-plugin-minify-mangle-names/__tests__/mangle-names-test.js
+++ b/packages/babel-plugin-minify-mangle-names/__tests__/mangle-names-test.js
@@ -177,7 +177,7 @@ describe("mangle-names", () => {
   });
 
   // https://phabricator.babeljs.io/T6957
-  xit("labels should not shadow bindings", () => {
+  it("labels should not shadow bindings", () => {
     const source = unpad(`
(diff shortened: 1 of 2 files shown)


kangax:
I submitted a proper fix so we'll try to get that in first — https://github.com/babel/babel/pull/4758

boopathi:
Boom !!!

PR #196, opened 2016-10-17
Implemented transform-regexp-constructors plugin

shinew:
Changes RegExp constructors to literals.

### DIFF
--- a/packages/babel-plugin-transform-regexp-constructors/.npmignore
+++ b/packages/babel-plugin-transform-regexp-constructors/.npmignore
@@ -1,0 +1,4 @@
+src
+__tests__
+node_modules
+*.log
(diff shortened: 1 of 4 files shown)


shinew:
Updated. Also, for RegExp constructors with 0 arguments, this would currently throw at `const pattern = evals[0].value;`. Catching this case at pseudo-compile time is not terrible, but should this behavior be changed/documented?

kangax:
Yeah, let's definitely add a check for 0 arguments. It should transform to `/(?:)/` (empty non-capturing group). Also, I would rename `evals` to `evaluatedArgs` for clarity.

shinew:
Updated to gracefully handle empty RegExp constructions.

kangax:
Thanks!

kangax:
@hzoo @loganfsmyth @danez do we want to also update babili preset?

hzoo:
yeah if want to add it now - we'll need the package.json for this package + adding it to the preset package.json if we do

kangax:
@shinew Let's do it. That way we could also see the impact of this plugin on performance with our benchmark.

Issue #25, opened 2016-06-01
More optimizations


kangax:
- [x] Remove unused args (e.g. `bar` in `function foo(bar) { }`)
- [x] Remove space after `return` when possible (e.g. `return [1,2,3]` &rarr; `return[1,2,3]`)
- [ ] `var result = expression(); foo.bar = result;` &rarr; `foo.bar = expression()`
- [ ] Better constant propagation (looks like it is in the dead-code-elimination plugin, but its a combination of both)

``` js
var b = () => {
  var x = 14;
  var y = 7 - x / 2;
  return y * (28 / x + 2);
};
```

``` js
var a = () => {
  var a = 14;

  return (7 - a / 2) * (28 / a + 2);
};
```

napa3um:
May be `undefined` -> `''.$`  instead `undefined` -> `void 0`?

bfred-it:
@napa3um Already suggested near the end of https://github.com/babel/babili/issues/25#issuecomment-230671685

@hzoo Would it make sense to open a repo for the suggestions as [suggested earlier](https://github.com/babel/babili/issues/25#issuecomment-230989103)? Or maybe just one-enhancement-per-issue so they can be easily discussed and tracked.

jokeyrhyme:
Came across this one, too: https://github.com/nolanlawson/optimize-js
Is there already a babel plugin for this?

kangax:
We should also turn `new Array` to `Array()` no matter which arguments since they're identical and there's no need for `new` /cc @shinew

bfred-it:
Doesn't that already happen? That was in my original implementation

kangax:
So strange, I could swear it wasn't happening for me in repl.

kangax:
Ouch, just noticed we don't even do `1000` → `1e3` and so on

pitaj:
I'd like to see unnecessary, single-use variables get inlined and also functions like this:
```js
(function ([, a, b]) {
  return (function (c, d) {
    return c + d;
  })(a, b);
})(thing);

(function ([, a, b]) {
  var e = (function (c, d) {
    return c + d;
  })(a, b);
  return e;
})(thing);

(function ([, a, b]) {
  (function (c, d) {
    log(c + d);
  })(a, b);
})(thing);

var a = b.something();
run(a.other());

// become

(function ([, c, d]) {
  return c + d;
})(thing);

(function ([, c, d]) {
  return c + d;
})(thing);

(function ([, c, d]) {
  log(c + d);
[...]

PR #200, opened 2016-10-20
Implemented babel-plugin-remove-unused-params.

shinew:
Removes unused trailing params for functions.

Note: not safe because it changes `length` as noted in https://github.com/babel/babili/issues/25.

### DIFF
--- a/packages/babel-plugin-remove-unused-params/README.md
+++ b/packages/babel-plugin-remove-unused-params/README.md
@@ -1,0 +1,53 @@
+# babel-plugin-remove-unused-params
+
+This plugin removes unused trailing function parameters.
+
+## Example
+
+**In**
+
+```javascript
+function foo(a, b) {
+  return a;
+}
+```
+
+**Out**
+
+```javascript
+function foo(a) {
+  return a;
+}
+```
+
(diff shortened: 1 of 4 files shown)


kangax:
Am I understanding correctly that for each param it will traverse scope inwards to see if it's referenced? If that's the case we might have perf problem (especially with large/deep functions) which could be fixed by traversing once and checking all params for each scope.

kangax:
Also I wonder what's a good way for us to start marking plugins as unsafe. Perhaps Boolean safe property on a plugin object? /@hzoo @loganfsmyth @danez

loganfsmyth:
Maybe make it clear in the README of the plugin, and only include them in the preset behind the `unsafe` option?

boopathi:
Isn't this already a part of DCE? What am I missing?

boopathi:
#182

kangax:
@boopathi ouch, I completely forgot about your PR and REPL had old version.

kangax:
I also marked it in our optimizations list as done. 

@shinew sorry for the confusion!

shinew:
no problem!

PR #210, opened 2016-10-20
Add option: topLevel to Mangler

### LINKED ISSUE #204: Doesn't mung some vars.
You can see [here](https://lodash.com/sw.js) that the `BUILD_REV` var stays unminified.

boopathi:
- (close #204)


packages/babel-plugin-minify-mangle-names/README.md
```diff
@@ -71,3 +71,4 @@ require("babel-core").transform("code", {
 + `blacklist` - A plain JS Object with keys as identifier names and values indicating whether to exclude
 + `eval` - mangle identifiers in scopes accessible by eval
 + `keepFnName` - prevent mangler from altering function names. Useful for code depending on `fn.name`
++ `topLevel` - mangle topLevel Identifiers
```
loganfsmyth: To me, it feels like this name should be something we generalize across plugins for standalone scripts, vs commonjs scripts. Thoughts? `topLevel` applies to the case of the mangler, but for example in the case we heard about on slack, it was the top-level behavior of DCE that was failing. `topLevel` could apply to both, but the name itself doesn't really say much about _why_ top-level bindings should/should not be treated as special cases.

How would you feel about this being `standaloneScript` with a default of `false`? So if people want to use Babili on a whole script file, rather than on individual CommonJS/ES6 modules, they'd have to opt into it?

kangax: Even though most of the options currently describe the behavior not the type (i.e. "do this optimization with this stuff" vs. "this code is of such type and will be minified accordingly, whatever that implies") I think what you're suggesting makes sense.

Can we just use `sourceType` then?

| sourceType | top level mangling | unreferenced vars immutable |
|---|---|---|
| module | yes | no |
| script | no | yes |

loganfsmyth: `sourceType` toggles parsing, but the problem is that a `sourceType: script` file could still be a CommonJS module (and thus able to be mangled at the top level), so I personally don't think we should use that flag to toggle this behavior.

packages/babel-plugin-minify-mangle-names/__tests__/mangle-names-test.js
```diff
@@ -983,4 +983,42 @@ describe("mangle-names", () => {
...
+      function a() {
+        if (b === "production") {
+          c.call();
+        }
+      }
+      const b = "production";
+      var c = function a() {
+        new d({
+          [b]: "foo",
+          a: a(c)
```
boopathi: Oops! This is wrong.

packages/babel-plugin-minify-mangle-names/README.md
boopathi: I'd still go with `topLevel` because it is simple and tells you what it does. I get the concerns that it doesn't tell you why toplevel bindings should/should not be treated as special cases. But I want to leave that to the user to decide. And here it's really upto the user to decide why he wants to use the option as it can be useful in multiple cases - whether he is parsing modules that he can choose to mangle toplevel, or whether his script is standalone that he can choose to mangle everything etc... In my view, the option is just to tell the user what it does rather than why you should do it.

loganfsmyth: I follow your reasoning for this mangling plugin, but it does worry my for the DCE plugin, since `topLevel` doesn't have much meaning in that case.

kangax: Would `sourceType: commonjs` be a good option then?

kangax: It just seems right to have a single source of truth when it comes to the way bindings are shared / can be mangled, etc.

loganfsmyth: @kangax Agreed: https://github.com/estree/estree/issues/136 :D

jdalton:
Does this need more eyes?

kangax:
Since we likely won't get this in ESTree anytime soon, let's just go with "topLevel", not to waste too much time on bikeshedding?

PR #216, opened 2016-10-22
Fix undefined in multiple sequence expressions (close #208)

### LINKED ISSUE #208: Empty block throws in certain cases
This currently throws but shouldn't:

```js
!function() {
  var x;
  { }
  alert(x);
}()
```

boopathi:


kangax:
Could you please add a test (as in #208) as well?

boopathi:
👍

PR #218, opened 2016-10-23
Add more tests for DCE'ing conditional expressions

erikdesjardins:
...which involve a couple more complex conditions than `true` or `false`

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -791,6 +791,30 @@ describe("dce-plugin", () => {
     expect(transform(source).trim()).toBe(expected);
   });
 
+  it("should evaluate conditional expressions 3", () => {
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -791,6 +791,30 @@ describe("dce-plugin", () => {
...
+  });
+
+  it("should evaluate conditional expressions 4", () => {
+    const source = "null ? a() : b();";
+    const expected = "b();";
+    expect(transform(source).trim()).toBe(expected);
+  });
+
+  it("should evaluate conditional expressions 5", () => {
+    const source = "'foo' === 'foo' ? a() : b();";
```
erikdesjardins: I wasn't sure if the assumption was that `'foo' === 'foo'` would be reduced to `true` by constant folding, but I didn't see any tests for that in the constant folding plugin.

So if you prefer I can move these two to `constant-folding-test.js`

kangax:
Thanks!

PR #219, opened 2016-10-24
Inline temp return

shinew:
This removes the temporary variable right before return statements, if it is also the return value.

### DIFF
--- a/packages/babel-plugin-transform-inline-temp-return/.npmignore
+++ b/packages/babel-plugin-transform-inline-temp-return/.npmignore
@@ -1,0 +1,4 @@
+src
+__tests__
+node_modules
+*.log
--- a/packages/babel-plugin-transform-inline-temp-return/README.md
+++ b/packages/babel-plugin-transform-inline-temp-return/README.md
@@ -1,0 +1,55 @@
+# babel-plugin-transform-inline-temp-return
+
+This removes the temporary variable right before return statements, if it is
+also the return value.
+
+## Example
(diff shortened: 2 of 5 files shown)


boopathi:
I'm sure single use variables were already replaced in deadcode elimination.

boopathi:
https://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=&code=function%20foo(a)%20%7B%0A%20%20let%20x%20%3D%20(b)%20%3D%3E%20a%20%2B%20b%3B%0A%20%20return%20x%0A%7D&experimental=true&loose=false&spec=false

shinew:
Right, but is there a reason why this case doesn't work? https://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=es2015%2Cstage-2&code=function%20foo()%20%7B%0A%20%20let%20x%20%3D%20bar()%3B%0A%20%20return%20x%3B%0A%7D&experimental=true&loose=false&spec=false

kangax:
We need to take care of it in DCE since that's where we already do this kind of optimizations. This is the line that prevents `foo()` from being inlined since it's reported as impure but something like `{ foo: "bar" }` is inlined fine — https://github.com/babel/babili/blob/master/packages/babel-plugin-minify-dead-code-elimination/src/index.js#L244-L246

kangax:
Closing since we'll be taking care of it in DCE

PR #220, opened 2016-10-25
Require blocks around single block-scope declaration

### LINKED ISSUE #213: var declaration inlined incorrectly in conditional
```js
if (x) {
  y;
} 
else {
  const z = foo(function() { z });
}
```

incorrectly minifies to:

```js
if(x)y;else const a=foo(function(){});
```

`else const ...`  is a SyntaxError

boopathi:
+ (close #213)

The var declaration removal should be done in DCE instead of simplify. This is necessary as simplify as a standalone plugin should not produce syntax errors.


kangax:
Will this work with `else { function x() { } ... }`?

boopathi:
`if(true) function x() {}` or `else function x() {}` is not syntax error.

boopathi:
It is in strict mode though.

PR #221, opened 2016-10-25
Improved temp variable inlining.

shinew:
Now inline impure rvals assigned to temp lvals if the lval is used in a "simple" way immediately after the assignment.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2164,4 +2164,82 @@ describe("dce-plugin", () => {
       }
     })).toBe(expected);
   });
+
+  it("should inline returns for temp variables", () => {
+    const source = unpad(`
+      function foo() {
+        let a = bar();
+        return a;
+      }
+    `);
+    const expected = unpad(`
+      function foo() {
+        return bar();
+      }
+    `);
+    expect(transform(source)).toBe(expected);
+  });
+
+  it("should inline calls with temp variables", () => {
+    const source = unpad(`
+      function foo() {
+        let a = f();
+        bar(a);
+      }
+    `);
+    const expected = unpad(`
+      function foo() {
+        bar(f());
+      }
+    `);
+    expect(transform(source)).toBe(expected);
+  });
+
+  it("should inline return calls with temp variables", () => {
+    const source = unpad(`
+      function foo() {
+        let a = f();
+        return bar(a);
+      }
+    `);
+    const expected = unpad(`
+      function foo() {
+        return bar(f());
+      }
+    `);
+    expect(transform(source)).toBe(expected);
+  });
+
+  it("should inline transitive temp variables", () => {
+    const source = unpad(`
+      function foo() {
+        let a = f();
+        let b = a;
+        return b;
+      }
+    `);
+    const expected = unpad(`
+      function foo() {
+        return f();
+      }
+    `);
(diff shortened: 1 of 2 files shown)


kangax:
There's couple issues here:

1. It seems to miss `var` (`var x = foo(); return x;` doesn't get inlined)
2. It does crude analysis which is easy to mess up: `var x = foo(); { }; return x;` is not inlined. Neither is (more realistic) `var x = foo(); if (smth) return x;`. Neither is `var x = foo(); bar(x, y)` (more than 1 args). And so on.

The way I see it, in an ideal implementation we should traverse a tree from node A (source) to node B (destination) checking if everything in between is pure.

Either that or we ditch this whole idea (due to complexity) and introduce something like "unsafeInlining" flag.

kangax:
@boopathi @hzoo what do you think we should do here? Try harder or be simpler?

boopathi:
I'm not sure. There are too many cases where you shouldn't change at all. If it's not hard to detect if we maintain the order of impure calls, we can try harder.

assumption: all function calls cause side-effects and order matters.

<details>
<summary>**Number of impure expressions**

If there are more than 1 impure expression, then optimize if and ONLY if the order is maintained - between and including the binding declaration and the last usage of the variable.
</summary>

```js
function foo() {
  let x = Foo(),
      y = Bar();
  return x;
}
function bar() {
  let x = Foo();
  return Bar(x); // order is maintained. So can replace.
}
function foo1() {
  let a = Foo(), b = Bar();
  return b + a; // order changes
}
function foo2() {
  let a = Foo();
  return Bar(Baz() + a); // order changes
}
```

</details>

<details>
<summary>**Async Callbacks**
</summary>

```js
function foo() {
  var x = foo();
  return new Promise((resolve, reject) => resolve(x));
}
function bar() {
  var x = foo();
  callAsync(() => x);
}
```

</details>

<details>
<summary>**Sync callbacks and Loops**
</summary>

```js
function foo() {
  let x = foo();
  return items.map(() => x);
}
function bar() {
  let x = foo();
  for (;;) {
    if (a) bar(x);
  }
}
```

</details>

<details>
<summary>**Side effecty getters and setters**

+ `pure_getters` as an option - where we need not count getters as impure expression
</summary>

```js
function foo() {
  let x = bar.foo();
  bar.baz(x);
}
function foo2() {
  let x = { foo: foo(), bar: bar() };
  module.exports = x; // order doesn't change. can replace
}
function foo3() {
  let x = foo();
  module.foo.exports = x; // order changes because of getter.
}
```

</details>

<details>
<summary>**Chain replacements**

If we are doing this, we should chain them
</summary>

```js
function baz() {
  let x = Foo(),
      y = Bar(x),
      z = Baz(y);
  return z; // order is maintained
}
```

</details>

<details>
<summary>**Destructuring**

Right now this is disabled in DCE. May be we can use the `pure_getters` option here as well
</summary>

```js
function foo() {
  let {x} = a();
  return x; // can be optimized to a().x
}
// but should bail when
function bar() {
  let {x, y} = a();
  return x;
}
```

Probably array destructuring will be shorter with a temp variable than using the iterator protocol ?

</details>

kangax:
Why should it bail on:

```js
function bar() {
  let {x, y} = a();
  return x;
}
```

If we ignore `a().x` being a potential getter, then we don't care about unreferenced `y` and could still reduce it to `return a().x`, no?

boopathi:
```js
let {x, y} = a(); // x and y getters are called here. If we remove y, it is not called.
```

boopathi:
Closing this as it's been inactive. Please feel free to open a new PR or leave a comment asking to reopen this PR. Thanks!

PR #224, opened 2016-10-26
Add test for labels shadow in mangle (close #185)

### LINKED ISSUE #185: Error mangling identifier in try-catch when there's same-named label

boopathi:


boopathi:
this passes on my machine after upgrade. fails for integration with block scope plugin. The 2 DCE tests are fixed in #225

boopathi:
I cleared cache for master and re-ran the CI, and now that test (integrate with block scoping plugin) is not failing anymore.

PR #225, opened 2016-10-26
Fix DCE tests with labels in separate namespace

boopathi:
This should work on babel version <6.18.0 and >6.18.0 ...

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -895,7 +895,13 @@ module.exports = ({ types: t, traverse }) => {
         // here we handle the break labels
         // if they are outside switch, we bail out
         // if they are within the case, we keep them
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -957,4 +963,16 @@ module.exports = ({ types: t, traverse }) => {
     return path.isFunctionDeclaration()
       || path.isVariableDeclaration({ kind: "var" });
   }
+
+  function getLabel(name, _path) {
+    let label, path = _path;
+    while (!path.isProgram()) {
```
hzoo: you can use path.find/path.findParent here right?

boopathi: yes...

boopathi: hmm.. But I want to return the label instead of the parent.

PR #230, opened 2016-10-27
Shorten assignments to itself

kangax:


packages/babel-plugin-minify-simplify/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "babel-plugin-minify-simplify",
-  "version": "0.0.3",
+  "version": "0.0.4",
```
boopathi: this should be done by lerna and not manually.

kangax: ok, removed

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,17 @@ module.exports = ({ types: t }) => {
   const or = (a, b) => t.logicalExpression("||", a, b);
   const and = (a, b) => t.logicalExpression("&&", a, b);
 
+  const operators = new Set([
```
kangax: We need to consider splitting plugins into multiple modules. This kind of mixing of helpers and supplemental stuff is not great ^. I think we could even strive for 1 type-of-transform per file.

packages/babel-plugin-minify-simplify/package.json
hzoo: don't need to change package.json either

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
...
 
+      AssignmentExpression(path) {
+
+        const right = path.get('right');
+        const left = path.get('left');
+
+        const canShorten = (
+          right.type === 'BinaryExpression' &&
+          operators.has(right.node.operator) &&
+          left.node.name === right.node.left.name
```
boopathi: requires `isIdentifier` check. if `left.node` and `right.node.left` are not identifiers, it can end up in `undefined === undefined` check and canShorten will be true - eg: `[a,b] = foo() + 1` - canShorten will be true for this.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
         ],
       },
 
+      AssignmentExpression(path) {
+
+        const right = path.get('right');
+        const left = path.get('left');
+
+        const canShorten = (
+          right.type === 'BinaryExpression' &&
```
boopathi: `right.isBinaryExpression()`

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
...
+        );
+
+		    if (!canShorten) return;
+
+        const canBeUpdateExpression = (
+          right.node.right.type === 'NumericLiteral' &&
+          right.node.right.value === 1 &&
+          updateOperators.has(right.node.operator));
+
+        if (left.node.name === undefined) {
```
boopathi: `left.isMemberExpression()` is better IMO, rather than depending on left not being an identifier.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
...
+          // TODO: transform MemberExpressions as well
+          // e.g. foo.bar = foo.bar + 123;
+          return;
+        }
+
+        if (canBeUpdateExpression) {
+
+          const newExpression = t.updateExpression(
+            right.node.operator + right.node.operator,
+            t.identifier(left.node.name));
```
boopathi: probably needs to handle member expression here ?

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
...
+
+          path.replaceWith(newExpression);
+        }
+        else {
+
+          const newExpression = t.assignmentExpression(
+            right.node.operator + '=',
+            t.identifier(left.node.name),
```
boopathi: handle memberExpressions

packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
```diff
@@ -2136,4 +2136,57 @@ describe("simplify-plugin", () => {
...
+      'foo += 123;',
+      'foo = 1 + foo;',
+      'x = x++;',
+      'foo = foo + bar + baz;',
+      'window.foo = window.foo + 1;',
+      'window.foo.bar = window.foo.baz + 1;',
+    ];
+
+    actual.forEach((test, index) => {
+      expect(transform(test)).toBe(expected[index]);
```
boopathi:
```js
expect(actual.map((s) => transform(s))).toEqual(expected);
```

here you compare two arrays and even if one fails in the list, the others still run. But when you do `forEach` when one of them fails, all else that follows won't be run at all.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +135,49 @@ module.exports = ({ types: t }) => {
...
+        const canShorten = (
+          right.type === 'BinaryExpression' &&
+          operators.has(right.node.operator) &&
+          left.node.name === right.node.left.name
+        );
+
+		    if (!canShorten) return;
+
+        const canBeUpdateExpression = (
+          right.node.right.type === 'NumericLiteral' &&
```
boopathi: `right.get("right").isNumericLiteral()` or `t.isNumericLiteral(right.node.right)`

packages/babel-plugin-minify-simplify/src/index.js
hzoo: Wonder if we can make a generic utility for handling both Identifier/MemberExpression?

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +180,55 @@ module.exports = ({ types: t }) => {
         ],
       },
 
+      AssignmentExpression(path) {
+
+        const rightExpr = path.get('right');
+        const leftExpr = path.get('left');
+
+        const canBeUpdateExpression = (
+          rightExpr.get('right').isNumericLiteral() &&
```
shinew: What's the behavior if `.get('right')` fails?

hzoo: Should be ok - gives a path anyway? http://astexplorer.net/#/qyjXEtbUxr

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+    '<<', '>>', '>>>',
+    '&', '|', '^', '/',
+    '**'
+  ]);
+
+  const updateOperators = new Set([
+    '+', '-'
+  ]);
+
+  function isEqual(arr1, arr2) {
```
shinew: This function should be `isEqualToPrefix` rather than `isEqual`. `isEqual([1], [1, 2]) === true` given this implementation.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+    const propNames = [getName(prop.node)];
+
+    while (obj.type !== 'Identifier' &&
+           obj.type !== 'ThisExpression' &&
+           obj.type !== 'Super') {
+      const node = obj.get('property').node;
+      if (node) {
+        propNames.push(getName(node));
+      }
+      obj = obj.get('object');
```
shinew: nit: `.get()` calls are more expensive than just directly referencing the nodes' children.

It's probably faster/clearer to use nodes instead of `.get()`, which returns paths. This doesn't need path-specific methods, like scoping.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+      return 'this';
+    }
+    if (node.type === 'Super') {
+      return 'super';
+    }
+    // augment identifiers so that they don't match
+    // string/number literals
+    // but still match against each other
+    return node.name
+      ? node.name + '_'
```
shinew: This seems a bit hacky -- would it make sense to return an object with "type" and "value" properties instead?

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+    if (!path.isMemberExpression()) {
+      return;
+    }
+
+    let obj = path.get('object');
+
+    const prop = path.get('property');
+    const propNames = [getName(prop.node)];
+
+    while (obj.type !== 'Identifier' &&
```
shinew: This is just to traverse the chain of MemberExpressions like `a.b.c.d`, right? If so, can't we just let the condition be `t.isMemberExpression(node)`?

packages/babel-plugin-minify-simplify/src/index.js
kangax: Ugh, good catch. So the reason it works is because property names are resolved RTL so `foo.bar` won't match `foo.bar.baz` due to arrays being `['bar', 'foo']` and `['baz', 'bar', 'foo']`. We can maybe do an additional length check which could be a nice compromise between speed and correctness?

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+  ]);
+
+  function isEqual(arr1, arr2) {
+    return arr1.every((value, index) => {
+      return value.toString() === arr2[index].toString();
+    });
+  }
+
+  function getName(node) {
+    if (node.type === 'ThisExpression') {
```
kangax: I don't really care; I would just want us to be consistent in the entire codebase. Do we want abstractions over performance or vice versa? Ditto with `.get(...)` vs node access. Let's chose one and stick to that?

packages/babel-plugin-minify-simplify/src/index.js
kangax: `get` was recommended by @boopathi so I tried to stick to that, but see response above. Let's just figure out which to use everywhere.

packages/babel-plugin-minify-simplify/src/index.js
kangax: I thought about object-based abstraction but wasn't sure about perf. impact of all the "unnecessary" intermediate objects. Definitely would be less hackish.

The reason I went with duck-typing instead of explicit `isIdentifier`/`isLiteral` is in case I was forgetting some other abstraction. I'm also not sure about assumption of if something is a literal it will have a value and so on.

I just checked and I see that `NullLiteral` doesn't have value. Neither does TemplateLiteral :/

`null` also throws since I call `toString` on it instead of `String(...)`. 

Let me clean this up a little.

packages/babel-plugin-minify-simplify/src/index.js
boopathi: It's also about being able to use type aliases. So `t.isSomething` is a better check that you don't have to worry about whether it's an alias or not. If we forget that it's an alias and add some code, it'll be buggy and hard to debug later.

packages/babel-plugin-minify-simplify/src/index.js
boopathi: it's useful to have path in many places especially when there is a bug... we can optimize to nodes later when we have it correct. path has more information and we might not know when we have a bug in the code. I faced issues where some information from path(at least parentPath) would be required to solve the issue.

packages/babel-plugin-minify-simplify/src/index.js
hzoo: Unless we start compiling `t.isThisExpression` -> `node.type === 'ThisExpression'` 😛 - I was saying we could write a few linting rules for these

kangax:
@boopathi @shinew fixed some things up, PTAL

kangax:
@boopathi I'd love to merge this ASAP and cut a new release. Could you please take one last look?

boopathi:
I'd try using it with other plugins - especially block-scoping, dce and simplify.

Mangler is in broken state. Let's fix that before a release.

shinew:
LGTM 👍

kangax:
I'm checking this addition on our codebase with all the other plugins.

Agreed about mangler.

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -124,6 +180,55 @@ module.exports = ({ types: t }) => {
...
+          rightExpr.get('right').node.value === 1 &&
+          updateOperators.has(rightExpr.node.operator));
+
+        if (leftExpr.isMemberExpression()) {
+
+          const leftPropNames = getPropNames(leftExpr);
+          const rightPropNames = getPropNames(rightExpr.get('left'));
+
+          if (!leftPropNames ||
+              leftPropNames.includes(undefined) ||
```
boopathi: `.indexOf` instead of `.includes` for node4

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -194,6 +250,57 @@ module.exports = ({ types: t }) => {
...
+              !rightPropNames ||
+              rightPropNames.includes(undefined) ||
```
boopathi: indexOf

packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -28,6 +28,62 @@ module.exports = ({ types: t }) => {
...
+    "<<", ">>", ">>>",
+    "&", "|", "^", "/",
+    "**"
+  ]);
+
+  const updateOperators = new Set([
+    "+", "-"
+  ]);
+
+  function isEqual(arr1, arr2) {
```
boopathi: this is specific to this feature, we should have a comment or change name to indicate that, since this is at the top level (nearly).

PR #237, opened 2016-10-31
update benchmarks for v0.0.8 [skip ci]

hzoo:
cc @iamdustan

A litte bit smaller

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,10 +157,10 @@ Add to your `.babelrc`'s plugins array.
 Backbone.js:
 ```
            raw     raw win gzip   gzip win parse time run
-uglify     21.79kB 221%    7.29kB 169%     1ms        247ms
-closure    21.67kB 223%    7.37kB 167%     2ms        1223ms
-babili     21.9kB  219%    7.46kB 164%     2ms        722ms
-closure js 24.01kB 191%    8.04kB 144%     2ms        3301ms
+uglify     21.79kB 221%    7.29kB 169%     2ms        320ms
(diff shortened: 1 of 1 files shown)


boopathi:
we need to find a way to update benchmarks on same hardware - run on CI server - probably not going to be same configuration container every time I suppose. How do projects generally run benchmarks ?

hzoo:
Oh another thing we can do is run a diff of the outputs

kangax:
We should definitely do a diff. That's what React does — compares local changes to master branch. As far as absolute numbers in readme I think it doesn't really matter. What matters is how they are in relation to gcc/uglify.

hzoo:
Oh yeah I think I saw an example of that before but forgot cc @zpao?

zpao:
We've not been running benchmarks per se but tallying some data so we can compare runs (I think we did some perf stuff in the past but weren't getting a ton of great data). For the tallying, we just update a text file as a lazy DB in a branch from CI, which lets us build http://isfiberreadyyet.com/. It's pretty boring actually :)

https://github.com/facebook/react/blob/master/.travis.yml#L86-L95
https://github.com/facebook/react/blob/master/scripts/facts-tracker/index.js
https://github.com/facebook/react/blob/facts/fiber-tests.txt

iamdustan:
that was fast!

kangax:
How are we so behind on jQuery comparing to uglify?! This looks very fishy.

hzoo:
Yeah we could also just run it without whitespace removal and do a diff if that can give us anything?

PR #241, opened 2016-11-02
fix for infinity as lval

### LINKED ISSUE #188: `Infinity = 1` blowing up simplify plugin
It shouldn't blow up; this is just a no-op operation but is otherwise completely acceptable.

vigneshshanmugam:
+  (close #188)


kangax:
Thanks!

Issue #226, opened 2016-10-26
Duplicate declaration in nested function that has a let variable with same name as function.


aickin:
Hey babili awesome folks. Thanks for all your work!

I think I've found a bug where babili throws a "Duplicate declaration" error when I have a nested named function whose name is the same as a let variable inside the function. 

Here's a reduced repro case:

http://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=&code=var%20sayHello%20%20%3D%20function(name)%20%7B%0A%20%20'use%20strict'%3B%0A%20%20%0A%20%20function%20foo(callback2)%20%7B%0A%20%20%20%20let%20foo%20%3D%20callback2()%3B%0A%20%20%20%20return%20%22hello%2C%20%22%20%2B%20foo%3B%0A%20%20%7D%0A%20%20%0A%20%20return%20foo(()%3D%3Ename)%3B%0A%7D&experimental=true&loose=false&spec=false&playground=false

This code snippet runs fine in latest Chrome, so I'm pretty sure it's valid ES2015. Let me know if you need any more info, and thanks again!

boopathi:
Thanks!

Hmm.. This is a tricky one.

bardiharborow:
I feel like this is an upstream Babel bug.

boopathi:
This is not a problem anymore in latest babili.

Slapbox:
@boopathi Using babili 0.1.2 I still got this error. Had to go and change variable names and it seems to work now. Not 100% sure, still waiting for it to process.

vigneshshanmugam:
@Slapbox can you give an example code?

Slapbox:
```
export function dateWithOffset() {
  let dateWithOffset;
  const date = new Date();
 	dateWithOffset = null;
  if (global.ntpOffset) {
    const dateMilliseconds = date.getTime() + global.ntpOffset;
    dateWithOffset = new Date(dateMilliseconds);
  } else {
    dateWithOffset = date;
  }
  return dateWithOffset;
}
```

vigneshshanmugam:
I tried with the 0.2.0 version and I cannot reproduce the error with your sample code.

Slapbox:
Ah I didn't realize there was a 0.2.0 under the rename. I'm still on 0.1.2. I assume the issue is resolved. I'll try 0.2.0 later and if I find otherwise I'll update you. Thanks very much.

PR #244, opened 2016-11-02
Ensure binding exists

kangax:
I ran into this with one of our files. Don't have an easily reproducible testcase but maybe someone can think of something?

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -840,7 +840,7 @@ module.exports = ({ types: t, traverse }) => {
     const binding = scope.getBinding(id.name);
 
     // Check if shadowed or is not referenced.
-    if (binding.path.node !== node || !binding.referenced) {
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -840,7 +840,7 @@ module.exports = ({ types: t, traverse }) => {
     const binding = scope.getBinding(id.name);
 
     // Check if shadowed or is not referenced.
-    if (binding.path.node !== node || !binding.referenced) {
+    if (binding && (binding.path.node !== node || !binding.referenced)) {
```
boopathi: this is for a function/class expression ... weird that there is no binding but an id exists.

kangax:
@boopathi so... you want me to add a comment that reads "This function is affected when fixing this bug too afaik - #226" Or something else? Sorry, I didn't understand.

boopathi:
I was essentially re-writing this function to fix #226 ... Never mind .. I'll just keep this in mind ..

PR #252, opened 2016-11-05
Fix mangler

boopathi:
(close #250)

<details>
<summary>Perf</summary>

```
[0] ~/workspace/minify (fix-mangler-3) > ./scripts/benchmark.js ./scripts/fixtures/jquery.js
testing babili
testing closure
testing closure js
testing uglify
           raw     raw win gzip    gzip win parse time run
uglify     94.27kB 218%    32.78kB 158%     7ms        1824ms
babili     94.61kB 217%    32.86kB 157%     8ms        4442ms
closure    94.23kB 218%    33.38kB 153%     9ms        7187ms
closure js 95.64kB 213%    33.78kB 150%     8ms        11115ms

pluginAlias                   time(ms) # visits time/visit(ms)
minify-simplify               3035.934 59849    0.051
minify-constant-folding       751.687  124240   0.006
minify-mangle-names           709.608  1        709.608
minify-guarded-expressions    300.972  11558    0.026
minify-dead-code-elimination  267.945  26       10.306
foreign$16                    88.646   56682    0.002
minify-infinity               85.528   56682    0.002
internal.shadowFunctions      85.188   58012    0.001
minify-type-constructors      71.767   7596     0.009
foreign$15                    57.087   4718     0.012
foreign$10                    29.604   17278    0.002
minify-numeric-literals       26.078   4315     0.006
foreign$12                    21.840   9251     0.002
minify-flip-comparisons       17.308   4718     0.004
transform-regexp-constructors 7.854    109      0.072
foreign$13                    7.286    1336     0.005
internal.blockHoist           5.560    2496     0.002
foreign$11                    4.553    878      0.005
minify-replace                0.139    1        0.139
```
</details>

Last Updated: 8c58a06

PR #261, opened 2016-11-08
Check binding exists

kangax:
Ran into this in one of our files when NOT using es2015 transform. I'll try to make a test case but don't have one for now.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -115,6 +115,8 @@ module.exports = ({ types: t, traverse }) => {
 
           if (param.isIdentifier()) {
             const binding = scope.bindings[param.node.name];
+            if (!binding) continue;
+
             if (binding.referenced) {
               // when the first binding is referenced (right to left)
               // exit without marking anything after this


boopathi:
Related: https://github.com/babel/babel/issues/4815 . So we should probably include a `scope.crawl` before running DCE. 

I guess a for statement using let in the init and a return statement inside the for loop might fail.

boopathi:
I had similar issues in mangler as well, where a BindingIdentifier has no binding - and I reduced it to this - https://github.com/babel/babili/pull/252/files#diff-18ba673c600d34ebd7daef11f67829f9R224 ... without [clearing the traverse cache and re-crawling the entire scope](https://github.com/babel/babili/pull/252/files#diff-18ba673c600d34ebd7daef11f67829f9R154) this will throw.

kangax:
@boopathi so what do you think we should do here? Drop this PR in favor of #252?

boopathi:
No. Suggesting we should do traverse.clearCache and program.scope.crawl() before anything in DCE, and check if that fixes the bug.

This check really is a work around where we have a BindingIdentifier(here a function parameter) and ignore it when we do not have a binding in the scope.

boopathi:
I didn't come across any other case where a binding is missing at obvious places. So, a snippet that reproduces this problem should be helpful to debug if we missed anything.

kangax:
I just ran into this again, while running over Nuclide code base. It seems to only happen when there's other plugins that transform code before Babili (https://github.com/facebook/nuclide/blob/master/pkg/nuclide-node-transpiler/lib/NodeTranspiler.js#L43-L75); transforming source code itself didn't exhibit the issue...

Issue #263, opened 2016-11-10
Regexp escaping needs to be more robust


kangax:
For example, `new RegExp('\\/')` still fails.

/cc @shinew

PR #264, opened 2016-11-11
RegExp literal fixes

shinew:
See https://github.com/babel/babili/issues/263.

There's a comment I made in the code about the need to prettify `let a = /	/;` into `let a = /\t\;`, which have the same behavior. This makes the generated literal more clear, at the cost of 1-3 extra characters per invisible newline character when minified. I'm not convinced this beautification is necessary since it's generated code anyway.

Thoughts?

### DIFF
--- a/packages/babel-plugin-transform-regexp-constructors/__tests__/transform-regexp-constructors-test.js
+++ b/packages/babel-plugin-transform-regexp-constructors/__tests__/transform-regexp-constructors-test.js
@@ -10,6 +10,18 @@ function transform(code) {
 }
 
 describe("transform-regexp-constructors-plugin", () => {
+  it("should work", () => {
+    const source = String.raw`var x = new RegExp('\\/');`;
+    const expected = String.raw`var x = /\//;`;
+    expect(transform(source)).toBe(expected);
+  });
+
+  it("should work 2", () => {
+    const source = String.raw`var x = new RegExp('\\n');`;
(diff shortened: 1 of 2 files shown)


kangax:
I think we shouldn't bother with "\t" similar to how we don't do this w. Babel literals and minified=true.

Does this work with backslash escapes?

```js
new RegExp('https?:\\/\\/(?:[\\dA-Za-z\\-]+\\.)+[\\dA-Za-z\\-]+(?::\\d+)?\\/[^\\s]*')
```

shinew:
Yeah, that case works I think.
Output string is: `"var x = /https?:\\/\\/(?:[\\dA-Za-z\\-]+\\.)+[\\dA-Za-z\\-]+(?::\\d+)?\\/[^\\s]*/;"`, corresponding to the code:`var x = /https?:\/\/(?:[\dA-Za-z\-]+\.)+[\dA-Za-z\-]+(?::\d+)?\/[^\s]*/;`. Running `'https://github.com/babel/babili/pull/264'.replace(x, '')` in the console returns `""`.

kangax:
Strange, I'm seeing test failure on travis — https://travis-ci.org/babel/babili/jobs/175548633 — but not locally

kangax:
Actually, no there's failures locally too. @shinew could you please take a look when you get a chance!

shinew:
My bad -- forgot to update test after restoring the '\t' literal.

PR #266, opened 2016-11-13
Fix if_return & dce ifelse..return within loops

### LINKED ISSUE #257: if..return brings functions into block
### Input 

```js
function foo() {
  bar();
  if(x) return;
  const {a}=b;
  function bar () {
    baz();
    bar();
  }
}
```

### Output 

```js
function foo() {
  if (bar(), !x) {
    function c() {
      baz(), c();
    }

    const { a: d } = b;c();
  }
}
```

[REPL](https://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=es2017&code=function%20foo()%20%7B%0A%20%20bar()%3B%0A%20%20if(x)%20return%3B%0A%20%20const%20%7Ba%7D%3Db%3B%0A%20%20function%20bar%20()%20%7B%0A%20%20%20%20baz()%3B%0A%20%20%20%20bar()%3B%0A%20%20%7D%0A%20%20bar()%3B%0A%7D&experimental=true&loose=false&spec=false)

### LINKED ISSUE #265: Infinite loop when simplify and deadcode used together
Input:

```js
function getParentConditionalPath(path) {
  let parentPath;
  while (parentPath = path.parentPath) {
    if (parentPath.isIfStatement() || parentPath.isConditionalExpression()) {
      if (path.key === "test") {
        return;
      } else {
        return parentPath;
      }
    } else {
      path = parentPath;
    }
  }
}
```

Output: 

```js
function getParentConditionalPath(path) {
  for (let parentPath; parentPath = path.parentPath;)
    if (!(parentPath.isIfStatement() || parentPath.isConditionalExpression()))
      path = parentPath;
    else if (path.key !== "test")
      return parentPath;
}
```

for `path.key === "test"` and the above condition fails, this causes an infinite loop where in the actual code, it simply returns undefined.

Reason:
Maybe because of if_return in simplify ... related #257

### LINKED ISSUE #275: minify-dead-code-elimination incorrectly removes return statement in a loop
Simplified repro:
```js
function remove(id) {
    for (var j = 0; j < array.length; ++j) {
        if (array[j] === id) {
            array.splice(j, 1);
            return;
        }
    }
}
```
Output:
```js
function remove(id) {
    for (var j = 0; j < array.length; ++j) {
        if (array[j] === id) {
            array.splice(j, 1);
        }
    }
}
```
In this case `return` is not dead code, since it terminates the loop. Interestingly `break` statement is not removed in the similar case.
[REPL link](https://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=&experimental=true&loose=true&spec=true&code=function%20remove(id)%20%7B%0A%20%20%20%20for%20(var%20j%20%3D%200%3B%20j%20%3C%20array.length%3B%20%2B%2Bj)%20%7B%0A%20%20%20%20%20%20%20%20if%20(array%5Bj%5D%20%3D%3D%3D%20id)%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20array.splice(j%2C%201)%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20return%3B%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%7D%0A)

boopathi:
+ (Close #265)
+ (Close #257)
+ (Close #275)

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2,11 +2,19 @@ jest.autoMockOff();
 
 const babel = require("babel-core");
 const unpad = require("../../../utils/unpad");
+const deadcode = require("../src/index");
+const simplify = require("../../babel-plugin-minify-simplify/src/index");
 
 function transform(code, options, babelOpts) {
   return babel.transform(code,  Object.assign({}, {
-    plugins: [[require("../src/index"), options]],
+    plugins: [[deadcode, options]],
   }, babelOpts)).code.trim();
+}
+
+function transformWithSimplify(code) {
+  return babel.transform(code, {
+    plugins: [deadcode, simplify]
+  }).code;
 }
 
 describe("dce-plugin", () => {
@@ -2209,6 +2217,7 @@ describe("dce-plugin", () => {
     `);
     expect(transform(source)).toBe(expected);
   });
+
   it("should NOT remove fn params for setters", () => {
     const source = unpad(`
       function foo() {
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2,13 +2,21 @@ jest.autoMockOff();
...
+const simplify = require("../../babel-plugin-minify-simplify/src/index");
 
 function transform(code, options, babelOpts) {
   return babel.transform(code,  Object.assign({}, {
-    plugins: [[require("../src/index"), options]],
+    plugins: [[deadcode, options]],
   }, babelOpts)).code.trim();
 }
 
+function transformWithSimplify(code) {
```
boopathi: I guess this is not necessary. Will remove it.

packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
```diff
@@ -573,10 +573,10 @@ describe("simplify-plugin", () => {
       }
       function bar(a) {
         if (!lol) try {
-          doThings();
-        } catch (e) {
-          doOtherThings();
-        }
+            doThings();
```
boopathi: I don't know why this happened.

boopathi:
Failing test: regexp-constructors.

PR #274, opened 2016-11-15
Fix version mismatches

boopathi:


kangax:
@boopathi @hzoo so only transform ones have their version matched with Babel, but minify ones will have their own?

hzoo:
It's not really matched with babel it's just that we published 6.8.0 before because they were existing packages while babili packages were starting at 0.0.0

PR #277, opened 2016-11-16
compile to node 4

hzoo:
This should be ok since we require node >=4? (map/set, etc)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,9 @@
 {
-  "presets": ["es2015"]
+  "presets": [
+    ["env", {
+      "targets": {
+        "node": "4"
+      }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,3 +1,9 @@
 {
-  "presets": ["es2015"]
+  "presets": [
+    ["env", {
+      "targets": {
+        "node": "4"
```
boopathi: I'm wondering why `4` is not a number and is a string.

hzoo: oops yeah should be a number

boopathi: Ah.. okay.. pushed to master.

PR #282, opened 2016-11-19
Don't transform earlyReturn on function enter

### LINKED ISSUE #281: Simplify drops function declaration for if_return
```js
function foo() {
  if (x)
    return;
  function bar() {}
  bar(a);
}
```

output: 

```js
function foo() {
  !x && bar(a), bar(a);
}
```

boopathi:
+ (close #281)


kangax:
This reminds me of our inefficient if `(!!foo) { ... }` being transformed to `!foo||log(1)`

PR #294, opened 2016-11-23
lift var declarations to initializer (fix #290)

### LINKED ISSUE #290: Lift var declarations to the loop initializer
Currently, this code:

```js
for (var i = 0; i < 0; i++) {
  var j = jj();
  console.log(j);
}
```
transforms to:
```js
for(var i=0;0>i;i++){var j=jj();console.log(j)}
```
but if we lift var declarations to the loop, we can get it down to:
```js
for(var i=0,j;0>i;i++){j=jj();console.log(j)}
```

This becomes irrelevant with multiple declarations so we might only consider doing this in certain cases but not others:

```js
for (var i = 0; i < 0; i++) {
  var j = jj(), k = kk(), l = ll();
  console.log(j, k, l);
}
```
transforms to:
```js
for(var i=0;0>i;i++){var j=jj(),k=kk(),l=ll();console.log(j,k,l)}
```
and is same length with lifted:
```js
for(var i=0,j,k,l;0>i;i++){j=jj(),k=kk(),l=ll();console.log(j,k,l)}
```

vigneshshanmugam:


vigneshshanmugam:
It breaks for this scenario.. 

```js
for (var i = 0; i < b; i++) {
  var j = 10, k = 11;
}
```

packages/babel-plugin-transform-merge-sibling-variables/src/index.js
```diff
@@ -1,9 +1,39 @@
...
     name: "transform-merge-sibling-variables",
     visitor: {
+      ForStatement(path) {
+        // Lift var declarations to the loop initializer
+        let body = path.get("body");
+        if (body.isBlockStatement()) {
+          body = body.get("body");
+          if (body[0].isVariableDeclaration({ kind: "var" })) {
+
+            if (body[1] && body[1].isVariableDeclaration({ kind: "var" })) {
```
boopathi: why is this required ?

```js
for(var k = 0;;) { var i = 0; var j = 0; }
```

packages/babel-plugin-transform-merge-sibling-variables/src/index.js
```diff
@@ -1,9 +1,39 @@
...
+            }
+
+            let firstNode = body[0].node.declarations[0];
+
+            if (!t.isIdentifier(firstNode.id)) {
+              return;
+            }
+
+            let init = path.get("init");
+            init.node.declarations = init.node.declarations.concat(
```
boopathi: will fail for `for(;;) {}`

vigneshshanmugam: can be fixed with body[0] && check..

boopathi:
```js
input: for(var i=foo;;){var j=bar;var k=baz}
opt 1: for(var i=foo;;){var j=bar,k=baz}
opt 2: for(var i=foo,j;;){j=bar;var k=baz}
```

We have to make sure that we end up with `opt 1` and not `opt 2` ...

vigneshshanmugam:
@boopathi addressed the cases you mentioned

boopathi:
I think we can lift the `let` declarations as well ...

vigneshshanmugam:
hmmm makes sense.. so var/let gets lifted both ways...

kangax:
Can we lift `const` too? Isn't is the same semantics as `let` in a loop?

vigneshshanmugam:
@kangax not sure about that..

packages/babel-plugin-transform-merge-sibling-variables/src/index.js
```diff
@@ -1,9 +1,51 @@
...
     visitor: {
+      ForStatement(path) {
+        // Lift declarations to the loop initializer
+        let body = path.get("body");
+        body = body.isBlockStatement() ? body.get("body") : [ body ];
+
+        liftDeclaration(body, "var");
+        liftDeclaration(body, "let");
+
+        function liftDeclaration(body, kind) {
```
boopathi: this function will be declared for every ForStatement, move it to an outer scope ?

vigneshshanmugam: oops yeah..

boopathi:
const in for loop cannot be lifted ... const requires Initialization or right hand side.

PR #303, opened 2016-11-26
Fix dce - recompute path & scope before pass

### LINKED ISSUE #301: Error w. JSX element in certain context
Getting this error:

_Property name of JSXOpeningElement expected node to be of a type ["JSXIdentifier","JSXMemberExpression"] but instead got "Identifier"_

for this code:

```js
async function f() {
  await Promise.all(
    async () => {
      const {x} = y;
      const X = x;
      return <X />;
    }
  )
}
```

Repl:

https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=true&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=async%20function%20f()%20%7B%0A%20%20await%20Promise.all(%0A%20%20%20%20async%20()%20%3D%3E%20%7B%0A%20%20%20%20%20%20const%20%7Bx%7D%20%3D%20y%3B%0A%20%20%20%20%20%20const%20X%20%3D%20x%3B%0A%20%20%20%20%20%20return%20%3CX%20%2F%3E%3B%0A%20%20%20%20%7D%0A%20%20)%0A%7D

boopathi:
+ (close #301)

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -699,6 +699,9 @@ module.exports = ({ types: t, traverse }) => {
             keepFnArgs = false,
           } = {}
         } = {}) {
+          traverse.clearCache();
+          path.scope.crawl();
+
           // We need to run this plugin in isolation.
           path.traverse(main, {
             functionToBindings: new Map(),


boopathi:
Should also address #261 IMO.

kangax:
Good idea

PR #304, opened 2016-11-26
Escape unicode newline in regex optimization

SimonSelg:
[In javascript \u2028 and \u2029 count as newlines too](http://www.ecma-international.org/ecma-262/5.1/#sec-7.3) (in addition to \u000a or \n and \u000d or \r).

Since only \r and \n are properly escaped this PR adds \u2028 and \u2029.


kangax:
Good catch, thanks!

PR #306, opened 2016-11-26
Pass per preset the babili plugins

boopathi:
This should fix most of the bugs that are caused by using it with es2015 preset or other plugins.


boopathi:
will fix the test cases if we decide to merge this ...

kangax:
I think this is a must since things are clearly broken without it.

@hzoo are you ok with this?

kangax:
@boopathi could you please fix the tests meanwhile

boopathi:
Done 👍 . It was simpler than I thought it'd be.

PR #319, opened 2016-12-04
Remove dollar from sh snippets

xtuc:
Since we have a clipboard functionality on the website, we don't want the `$` anymore.

Also did in babel repo babel/babel#4910

Fixes babel/babel.github.io#1029


boopathi:
Cool. Thanks!

PR #325, opened 2016-12-05
remove comments by default in babel-preset-babili

hzoo:
Then we don't have to modify the readme unless we want to say somewhere it also removes comments.

This can be a bug fix (it was supposed to remove comments) or breaking change (it wasn't removing it before)

- [x] test

### DIFF
--- a/packages/babel-preset-babili/__tests__/preset-tests.js
+++ b/packages/babel-preset-babili/__tests__/preset-tests.js
@@ -39,4 +39,26 @@ describe("preset", () => {
     `);
     expect(transform(source)).toBe(expected);
   });
+
+  it("should fix remove comments", () => {
+    const source = unpad(`
+      var asdf = 1; // test
+    `);
+    const expected = unpad(`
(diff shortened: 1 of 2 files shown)


boopathi:
I think the default should be to preserve `@preserve` and `@license` comments

hzoo:
Oh is that not the default?

boopathi:
Ah. Cool. I was thinking `comments: false` will remove all comments.

activedecay:
> Then we don't have to modify the readme unless we want to say somewhere it also removes comments.

well, i'm using your package, but i don't know how to remove comments, any help would be appreciated.

```
var gulp = require('gulp')
var babel = require('gulp-babel')
var fs = require('fs');

exports.task = function () {
    return gulp.src('e2e/test/pages/**.js')
        .pipe(babel({presets: ['babili']}))
        // i want to remove comments here
        .pipe(gulp.dest('e2e/lib'))
};

```

PR #335, opened 2016-12-09
Add fb package and an option to bench local file

kangax:


kangax:
@bmaurer could you please check the license headers. Do I need to fill into "20__"?

bmaurer:
yeah, you need to fill in the date. also we should put the FB files (and the license file) in a separate directory so it's clear what the scope of the example license is

PR #340, opened 2016-12-13
don't simplify assignment when it is not an equal operator (fix #339)

### LINKED ISSUE #339: simplify assignment doesn't check if it is a shorthand equal
Hi, currently with babel-plugin-minify-simplify at version 0.0.5,

```javascript
x = x + (x >> 8);
```
is "simplified" to :

```javascript
x >>= 8;
```
in two steps with this in between : 
```javascript
x += x >> 8;
```
So it doesn't check if the assignment is a shorthand operator.

It affects [Immutable.js in the popCount method](https://github.com/facebook/immutable-js/blob/c65062599ecad54a1ad5bacd72f65e8f9ef7449b/src/Map.js#L809).

amarant:


kangax:
Looks good, thanks!

PR #344, opened 2016-12-16
Inline value where MemberExpression.object is an Expression.

### LINKED ISSUE #336: MemberExpressions where .object is an ObjectExpression should have the relevant property inlined.
Input:

```javascript
function foo() {
  const bar = {
    a: 1,
    b: 2
  };
  console.log(bar.b)
}
```

Output:

```javascript
function foo(){console.log({a:1,b:2}.b)}
```

Expected Output:

```javascript
function foo(){console.log(1)}
```

(Part of a series of issues that affect the optimal minification of twbs/bootstrap.)

bardiharborow:
Heavily WIP. Fixes #336. Credit to @kovensky for help.

### DIFF
--- a/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
+++ b/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
@@ -2600,6 +2600,29 @@ describe("simplify-plugin", () => {
     expect(transform(source)).toBe(expected);
   });
 
+  it("should replace object literal member expressions with values", () => {
+    const source = unpad(`
+      foo = {
+        a: 1,
+        b: 2,
+        c: 3
+      }.b;
+    `);
+    const expected = unpad(`
+      foo = 2;
+    `);
+    expect(transform(source)).toBe(expected);
+  });
+
+  it("should not replace object literal member expressions assignments", () => {
+    const source = unpad(`
+      foo({
+        bar: 1
+      }.bar = 2);
+    `);
+    expect(transform(source)).toBe(source);
+  });
+
   it("should fix issue#281 with if..return", () => {
     const source = unpad(`
       function foo() {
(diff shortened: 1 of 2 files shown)


kangax:
I'm thinking the algorithm, roughly, should be:

1. Go over all member expressions
2. Check if we can resolve their value (e.g. the base object — `foo` in `foo.bar` — is in scope, is an object and has a "bar" property that is a pure value)
3. Check if we can inline it (i.e. the base object is not used anywhere else EXCEPT in other member expressions, since we might inline it there too)
4. Inline it (i.e. replace with a value and remove original assignment)

Some of the cases we need to cover, off the top of my head:

```js
var foo = { x : 1, y: 2 };
foo.x;
```
becomes `1`

```js
var foo = { x: 1, y: 2 };
foo.x + foo.y;
```
becomes `1 + 2` (and could then be inlined further by other plugins). This is the case of base object being referenced few times.

```js
var foo = { x: 1, y: bar() }
foo.x;
```
Stays the same (or perhaps we inline but leave original assignment) because assignment has side effects (`bar()` invocation).

```js
var foo = { bar:  { x: 1 } }
foo.bar.x;
```
becomes `1` (meaning that we need to be able to work with multiple levels)

```js
var foo = { x: foo, y: function() { ... } };
foo.x;
```
stays the same due to circular reference.

We should probably check Uglify and/or GCC as they might have more test cases.

bardiharborow:
This PR tackles the issue of `console.log({ x : 1, y: 2 }.x)` and relies on `dead-code-elimination` to inline the ObjectExpression in the first place. I think what you are suggesting is to inline the value in the directly in `dead-code-elimination` rather this two-step process.

kangax:
@bardiharborow IMO, this is similar to minify-constant-folding so it could be a separate plugin that inlines/folds object values instead.

bardiharborow:
This PR is now quite old, and likely doesn't implement an optimal algorithm anyway, so closing out.

PR #371, opened 2017-01-10
Add mark eval scopes helper and deopt DCE fn unused params

### LINKED ISSUE #366: babili is removing function arguments in scope of eval call
The following code:

```js
'use strict';
function myEval(code, _var_) {
  eval(code);
}
myEval('console.log(_var_)', 'myValue');
```

Is transformed to:
```js
'use strict';function myEval(code){eval(code)}myEval('console.log(_var_)','myValue');
```

Which does not work.

I'm using `babel-preset-babili` v0.0.9. The bug was not present in v0.0.5. I suppose it was introduced in https://github.com/babel/babili/commit/10b114f26533a58f66bb8a8ae90335c7353d1085.

### LINKED ISSUE #367: babili is removing variables in scope of eval call
Related to https://github.com/babel/babili/issues/366 but it might be a different issue because reverting to v0.0.5 does not fix the issue.

Input:
```js
'use strict';
function createFilters(filterDefinitions) {
  var filters = {};
  for (var i = 0; i < filterDefinitions.length; i++) {
    eval('filters[filterDefinitions[i].name] = function(value) { ' + filterDefinitions[i].script + '}');
  }
  return filters;
}
createFilters([
  {name: 'identity', script: 'return value;'},
  {name: 'add 42', script: 'return value + 42;'}
]);
```

Output:
```js
'use strict';function createFilters(filterDefinitions){for(var i=0;i<filterDefinitions.length;i++)eval('filters[filterDefinitions[i].name] = function(value) { '+filterDefinitions[i].script+'}');return{}}createFilters([{name:'identity',script:'return value;'},{name:'add 42',script:'return value + 42;'}]);
```

boopathi:
+ (fixes #366)
+ (fixes #367)


boopathi:
The helper attaches to the scope object directly instead of returning the list of scopes that contain eval. This is to avoid traversing the program twice - mangler, dce.

boopathi:
Also, this now only deopts eval scopes for only a few transforms under the visitor `Scope`. Have to look into other places in DCE that require deopt for eval.

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2306,4 +2306,30 @@ describe("dce-plugin", () => {
...
+        return b;
+      }
+      function b(c, d, e) {
+        (1, eval)(";");
+        return c;
+      }
+    `);
+
+    const expected = unpad(`
+      function a(b, c, d) {
```
kangax: Will this work with nested functions too? i.e. making sure it doesn't mangle top-level vars either.

packages/babel-helper-mark-eval-scopes/src/index.js
```diff
@@ -0,0 +1,52 @@
+"use strict";
+
+const EVAL_SCOPE_MARKER = "evalInScope";
```
kangax: Should we use Symbol instead?

boopathi: yes. Will change. I was using `scope.evalInScope` directly.

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
boopathi: added more tests

boopathi:
> I feel like this is something that should be done on Babel level

I think so too, but if we keep it in babel, it should be CORRECT rather than optimised for a few use cases. Like now, we don't recompute between multiple calls of `hasEval()`/`isUnsafeScope()` - as we assume that we don't create new scopes. If we do, we should manually recompute these things by calling `markEvalScopes()` again. 

If we implement this in babel, every call of `hasEval`/`isUnsafeScope` should trigger `scope.traverse` to find if it contains eval. And there are other cases like - eval is removed.

kangax:
Yeah, this brings us again to the usual issues of keeping track of proper state and revalidation. Just like we don't update bindings or certain path information right now after changing nodes.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -710,6 +715,10 @@ module.exports = ({ types: t, traverse }) => {
           traverse.clearCache();
           path.scope.crawl();
 
+          if (!isEvalScopesMarked(path.scope)) {
+            markEvalScopes(path);
```
kangax: I'd love to merge this for now.

Just one suggestion — let's move `isEvalScopesMarked` into `markEvalScopes` (checking it within and returning if scope is already marked). That way API becomes simpler (`markEvalScopes(path)`), there's less vars to exports, etc.

Another thing to consider is that `markEvalScopes` operates on path, and `hasEval` operates on scope. In an ideal world, that means these methods belong to those objects — `path.markScopesWithEval`, `scope.hasDirectEvals`.

boopathi: I didn't put isMarked inside markEvalScopes because if something is changed after marking - we need to remark.

boopathi: > that means these methods belong to those objects — path.markScopesWithEval, scope.hasDirectEvals

Yes!!

kangax: Do you want to implement them in Babel instead?

boopathi: /cc @hzoo what do you think ?

kangax:
Merging for now. Later, we can implement this in Babel instead.

kangax:
@boopathi thanks for a speedy fix!

PR #384, opened 2017-01-20
Escape "</script" when inlining strings

### LINKED ISSUE #382: Allow for exceptions in string concat inlining
Originally brought up by @spicyj.

Since `'</' + 'script'` is a pretty common way of avoiding "</script" strings inside script tags, we need to figure out how to allow for cases like this.

Alternatively, we can replace "</" sequence with unicode representation (`\u003c\u002f`) but I'm not sure if it's minifier's job to do this.

One immediate solution I see is to add an option of blacklisting concatenable strings:

```js
babel.transform(code, { presets: [['babili', { concatStringsBlacklist: '</', 'script' }]] });
```

It's unclear which logic to use for a list of blacklisted strings — when ALL match or when ANY match? If we use ANY then we could be missing out on concatenating "</div", "</some:component", etc.

If we were to go with escaping:

```js
babel.transform(code, { presets: [['babili', { unicodeStrings: ['</'] }]] });
```

But then it feels like something Babel (Babel's printer, actually) should be taking care of.

I'm curious to hear what @mathiasbynens thinks.

/cc @hzoo

kangax:
Fixes #382

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/src/index.js
+++ b/packages/babel-plugin-minify-constant-folding/src/index.js
@@ -113,9 +113,9 @@ module.exports = ({ types: t, traverse }) => {
           }
 
           // https://github.com/babel/babili/issues/382
-          if (typeof res.value === 'string' &&
-              res.value.indexOf('</script') > -1) {
-            res.value = res.value.replace(/<\/script/g, '<\\/script');
+          if (typeof res.value === "string" &&
+              res.value.indexOf("</script") > -1) {
+            res.value = res.value.replace(/<\/script/g, "<\\/script");
           }
 
           const node = t.valueToNode(res.value);
--- a/packages/babel-plugin-minify-constant-folding/src/index.js
+++ b/packages/babel-plugin-minify-constant-folding/src/index.js
@@ -112,6 +112,12 @@ module.exports = ({ types: t, traverse }) => {
             }
           }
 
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-constant-folding/src/index.js
```diff
@@ -112,6 +112,12 @@ module.exports = ({ types: t, traverse }) => {
             }
           }
 
+          // https://github.com/babel/babili/issues/382
+          if (typeof res.value === "string" &&
+              res.value.indexOf("</script") > -1) {
```
vigneshshanmugam: Can we use `includes` instead of `indexof` since we are already targeting node 4?

targos: `includes` was introduced in node 6

goto-bus-stop:
should this also account for the `<!--` case mentioned in https://github.com/babel/babili/issues/382#issuecomment-273873585 ?

packages/babel-plugin-minify-constant-folding/src/index.js
vigneshshanmugam: Yeah...I know that..But we are transpiling the code to node 4.. So it shouldnt be a problem

kangax:
We could just use `jsesc` then. Especially since it's already used by Babel.

PR #386, opened 2017-01-22
DCE: Deopt impure statements in If.test

### LINKED ISSUE #385: Incorrect transformation on React project with both simplify and dead code elimination enabled
I couldn't make a minimal repro, but compiling my project with these two plugins enabled leads to the following error message (it's worth mentioning that babili is run on single fairly large file created by Webpack):

```
Uncaught TypeError: inst.render is not a function        ReactCompositeComponent.js:811
``` 

The issue doesn't manifest with either of the plugins disabled. I narrowed the bug to `createPrevExpressionEater("if")` in `simplify` whose result is removed by the `if (evaluateTest === false)` block in the `IfStatement` part of `dead-code-elimination`. 

If there's any diagnostic I can run to be more helpful I'd be happy to do so.

boopathi:
TODO:

+ [x] Test case for impure statements
+ [x] Preset test cases with other plugins ?

### Note:
**This moves IfStatement from program.traverse in DCE to DCEPlugin.visitor.**

+ (Fix #385)
+ Add isPure check to IfStatement visitor in DCE
+ Move IfStatement visitor from program.traverse to DCEPlugin.visitor - so that it executes for all IfStatements as babel traverse happens and NOT during the single pass within programPath.traverse in DCE. This eliminates the need for a second pass for things like the one mentioned in #385.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2421,4 +2421,13 @@ describe("dce-plugin", () => {
 
     expect(transform(source)).toBe(expected);
   });
+
+  it("should deopt impure expressions in if statements", () => {
+    const source = unpad(`
+      if (a.b(), true) {
+        foo();
+      }
+    `);
+    expect(transform(source)).toBe(source);
+  });
 });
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2421,4 +2421,13 @@ describe("dce-plugin", () => {
...
     expect(transform(source)).toBe(expected);
   });
+
+  it("should deopt impure expressions in if statements", () => {
+    const source = unpad(`
+      if (a.b(), true) {
+        foo();
+      }
+    `);
+    expect(transform(source)).toBe(source);
```
boopathi: this can be improved.

boopathi:
@kangax the failing tests are - getting `</script` instead of `<\\\\/script`

PR #387, opened 2017-01-23
CallExpressions to TaggedTemplateLiterals

boopathi:
+ Unsafe Transformation.
+ Adds option `simplify.callsToTags`
+ Adds option `ecmaVersion` (Default = 2015) to preset - now passed only to simplify. Can be used for other transformations later
+ Depends on https://github.com/babel/babel/issues/5192

Should we have this in some other plugin ? It can be present anywhere - nothing related to simplify (considering simplify already has too many transformations)

### DIFF
--- a/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
+++ b/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
@@ -37,9 +37,9 @@ const plugin = require("../src/index");
 const comparisonPlugin = require("../../babel-plugin-transform-simplify-comparison-operators/src");
 const unpad = require("../../../utils/unpad");
 
-function transform(code) {
+function transform(code, options) {
   return babel.transform(code,  {
-    plugins: [plugin],
+    plugins: [ [plugin, options] ],
   }).code;
 }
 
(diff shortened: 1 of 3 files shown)


kangax:
Maybe we should call an option "es2015=true", consistent with es2015 presets in babel, etc. 

Other than that, this addition makes sense. I have a feeling it's not very impactful due to gzip, and possibly even worse, but we'd need to see benchmarks for that.

/cc @hzoo

hzoo:
Yeah I'm not sure about `ecmaVersion` (I don't know a good name for it either) but I guess we just need to signal that it means that is the minimum supported target version.

Future thoughts: would be better to target a specific feature though (not sure how it would work). An integration with babel-preset-env would be pretty great then since we could pass down the data regarding TaggedTemplate support -> this plugin to turn it on or not.

boopathi:
Oh. Yeah. I'll change it to use the preset-env data and implement option as `target` like in `babel-preset-env` instead of `ecmaVersion`. This will be useful for many other transformations as well - like we can convert some es5 to es2015 if it is smaller.

kzc:
Maybe I'm missing something, but is this proposed transform sound?
```
$ node -e 'function foo(x){console.log(JSON.stringify(x));} foo("bar"); foo`bar`;'
"bar"
["bar"]
```
```
$ node -v
v6.9.0
```

boopathi:
@kzc Good catch. I forgot about that. Thanks!

Closing this. Will try other ES2015+ transformations with the options discussed above.

PR #391, opened 2017-01-26
Inline hoisted, post-return declarations properly

### LINKED ISSUE #192: Incorrect inlining of unreachable statement
``` js
var x = 0;
function f1(){
  function f2(){
    return x;
  };
  return f2();
  var x = 1;
}
```

minifies to:

``` js
var x=0;function f1(){;return function(){return 1}()}
```

Notice inlined `x=1` which shouldn't be the case since the assignment is after return statement and so it's never executed. However `var x` still declares variable so the inlined value should be `undefined` or nothing at all. 

Ideally, the entire f1 body should be eliminated via DCE.

kangax:
Fixes #192

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2478,4 +2478,29 @@ describe("dce-plugin", () => {
 
     expect(transform(source)).toBe(expected);
   });
+
+  it("should not remove vars after return statement #2", () => {
+
+    const source = unpad(`
+      var x = 0;
+      function f1(){
+        function f2(){
+          return x;
+        };
+        return f2();
+        var x = 1;
+      }
+    `);
+
+    const expected = unpad(`
+      var x = 0;
+      function f1() {
+        return function () {
+          return undefined;
+        }();
+      }
+    `);
+
+    expect(transform(source)).toBe(expected);
+  });
 });
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2503,4 +2503,22 @@ describe("dce-plugin", () => {
 
     expect(transform(source)).toBe(expected);
   });
+
+  it("should not remove vars after return statement #3", () => {
+
+    const source = unpad(`
+      function foo() {
+        bar = x;
+        var x = 1;
+      }
+    `);
+
+    const expected = unpad(`
+      function foo() {
+        bar = undefined;
+      }
+    `);
+
+    expect(transform(source)).toBe(expected);
+  });
 });
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2496,8 +2496,9 @@ describe("dce-plugin", () => {
       var x = 0;
       function f1() {
         return function () {
-          return void 0;
+          return x;
         }();
+        var x = 1;
       }
     `);
 
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2460,4 +2460,22 @@ describe("dce-plugin", () => {
     `);
     expect(transform(source)).toBe(expected);
   });
+
+  it("should not remove vars after return statement", () => {
+
+    const source = unpad(`
+      function f() {
+        return x;
+        var x = 1;
+      }
+    `);
+
+    const expected = unpad(`
+      function f() {
+        return undefined;
+      }
+    `);
+
+    expect(transform(source)).toBe(expected);
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2460,4 +2460,47 @@ describe("dce-plugin", () => {
...
+        function f2(){
+          return x;
+        };
+        return f2();
+        var x = 1;
+      }
+    `);
+
+    const expected = unpad(`
+      var x = 0;
```
kangax: This is not ideal, of course. But we could revisit later.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -3,6 +3,15 @@
 const some = require("lodash.some");
 const { markEvalScopes, isMarked: isEvalScopesMarked, hasEval } = require("babel-helper-mark-eval-scopes");
 
+function forEachPrevSibling(path, callback) {
```
kangax: Considering that DCE has a lot of cases of `path.parentPath.getSibling(path.parentPath.key - 1)` and `path.parentPath.getSibling(path.parentPath.key + 1)`, I think it makes sense to add helper traversal methods to Babel.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -238,10 +247,25 @@ module.exports = ({ types: t, traverse }) => {
...
+
               let replacement = binding.path.node;
               let replacementPath = binding.path;
+              let isAfterReturn = false;
+
               if (t.isVariableDeclarator(replacement)) {
-                replacement = replacement.init;
+
+                forEachPrevSibling(replacementPath, (path) => {
+                  if (t.isReturnStatement(path)) {
```
boopathi: this will fail for cases like - 

```js
function foo() {
  bar = x;
  var x = 1;
}
```

packages/babel-plugin-minify-dead-code-elimination/src/index.js
kangax: Filed https://github.com/babel/babel/issues/5223

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -238,10 +247,25 @@ module.exports = ({ types: t, traverse }) => {
...
+                    isAfterReturn = true;
+                  }
+                });
+
+                // simulate hoisting by replacing value
+                // with undefined if declaration is after return
+                replacement = isAfterReturn
+                  ? t.identifier("undefined")
```
boopathi: `void 0` is safe. We can't assume that there is no local variable called `undefined`

packages/babel-plugin-minify-dead-code-elimination/src/index.js
kangax: You're right. But then we have a much "deeper" problem:

```js
function foo() {
  (function() {
    alert(function(){ return x }());
  })();
  var x = 1;
}
```

This brings us again to CFG and knowing when an identifier is referenced before declaration (having to traverse entire tree before declaration).

I think in this case it would be simpler to just hoist vars before doing DCE. So this:

```js
function foo() {
  bar = x;
  var x = 1;
}
```

would first become:

```js
function foo() {
  var x;
  bar = x;
  x = 1;
}
```

which could then be transformed to:

```js
function foo() {
  bar = undefined;
}
```

packages/babel-plugin-minify-dead-code-elimination/src/index.js
kangax: Good point, thanks

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -238,10 +247,25 @@ module.exports = ({ types: t, traverse }) => {
...
+                  : replacement.init;
+
```
boopathi: a few lines below this, `replacementPath = replacementPath.get("init")` - we might have to fix this here too. Else replacement and replacementPath might point to different things.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
boopathi: Case 1 can be solved without a CFG ... `siblingPath.isAncestor(referencedPath)` will tell you this.

```js
function foo() {
  (() => {
    return x // x - ReferencedPath
  })() // CallExpression SiblingPath
  var x = 1;
}
```

kangax: Right, for each binding reference we can traverse a subtree upwards and check if parent is a previous sibling. We also need to check that identifier is not already defined:

```js
function foo() {
  ((x) => {
    return x // x - ReferencedPath but is a local var
  })() // CallExpression SiblingPath
  var x = 1;
}
```

boopathi: Yes! But, this case we are dealing with is "one-time-used-constant" - binding.references === 1 and binding.constant, and we already have the path to the reference (`binding.referencePaths[0]`). So the local variable case wouldn't be this referencePath.

kangax:
Ok, this was easier than I thought. And your case worked right away. What should we replace init path with?

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
boopathi: It's actually correct. This is not removed because it's a global.

boopathi: Created #392

boopathi:
One more case - I think we should deopt if the reference is in a different scope - instead of replacing it with undefined.

```js
function foo() {
  Promise.resolve().then(() => console.log(x));
  var x = 1;
}
```

will log `1` but we will optimize it to `undefined`

kangax:
Good point. We don't know if other scopes are executed (a)synchronously. Deopted.

PR #410, opened 2017-02-08
Implement minify-builtins plugin

### LINKED ISSUE #363: Create aliases for built in functions
I mean, if I use multiple times `Math.floor`, it would make sense to assign it to a variable and use the variable. 

```
function(x, y,z) {
  return { Math.floor(x), Math.floor(y), Math.floor(z)  };
}
```

To

```
function(x, y, z) {
  let a = Math.floor;
  return { a(x), a(y), a(z) };
}
```

vigneshshanmugam:
fixes #363


boopathi:
Thoughts: I think we could also help constant-propagation by evaluating the expressions if the arguments are constants ?

```js
let a = 1.8, b = 2.5;
let x = Math.floor(Math.max(a, b));
foo(x);
```

could be transformed to

```js
foo(2);
```

vigneshshanmugam:
@boopathi I like the idea in general, I believe it shouldn't cause any issues.. 

Shall i create a new issue for that and fix it in a new PR?

boopathi:
Part of this PR would be great.

vigneshshanmugam:
Cool then.. will give it a try..

vigneshshanmugam:
Handled the optimisations.

packages/babel-plugin-transform-built-ins/src/builtins.js
```diff
@@ -0,0 +1,79 @@
+module.exports = {
```
boopathi: The way babel does this is to maintain only `[String, Number, Math]` 

https://github.com/babel/babel/blob/f8ffe03/packages/babel-traverse/src/path/evaluation.js#L5 

So, I think it's relatively safe to assume that `Math.x` will be a side-effect free evaluation except `Math.random`

packages/babel-plugin-transform-built-ins/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+        CallExpression: {
+          exit(path) {
+            const callee = path.get("callee");
+            if (!callee.isMemberExpression()) {
+              return;
+            }
+
+            const expName = getExpressionName(callee);
+            if (isBuiltin(expName)) {
+              const result = evaluate(path);
```
boopathi: we have to deopt when we have side-effecty arguments 

for example,

```js
Math.floor((foo(), 1));
```

packages/babel-plugin-transform-built-ins/README.md
```diff
@@ -0,0 +1,51 @@
+# babel-plugin-transform-built-ins
```
boopathi: should we just call it `babel-plugin-minify-builtins`

packages/babel-plugin-transform-built-ins/src/index.js
```diff
@@ -0,0 +1,128 @@
+"use strict";
+
+const { methods, properties } = require("./builtins");
+const evaluate = require("babel-helper-evaluate-path");
+
+
+
```
boopathi: empty lines

packages/babel-plugin-transform-built-ins/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+        },
+        CallExpression: {
+          exit(path) {
+            const callee = path.get("callee");
+            if (!callee.isMemberExpression()) {
+              return;
+            }
+
+            const expName = getExpressionName(callee);
+            if (isBuiltin(expName)) {
```
boopathi: should check if it's a computed property

Will fail for `Math[max]()` will assume it as `Math.max()`.

packages/babel-plugin-transform-built-ins/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+    },
+  };
+
+  function memberToString(memberExpr) {
+    const {object, property} = memberExpr.node;
+    let result = "";
+
+    if (t.isIdentifier(object)) result += object.name;
+    if (t.isMemberExpression(object)) result += memberToString(object);
+    if (t.isIdentifier(property)) result += property.name;
```
boopathi: should check for computed properties. same as above.

packages/babel-plugin-transform-built-ins/src/builtins.js
vigneshshanmugam: Ahh good one.. What about Date.x?

packages/babel-plugin-transform-built-ins/README.md
vigneshshanmugam: Hmmm.. dono really..

packages/babel-plugin-transform-built-ins/src/builtins.js
boopathi: We can chuck the Date for now.

packages/babel-plugin-transform-built-ins/README.md
boopathi: Let's just change it to `babel-plugin-minify-builtins`

packages/babel-plugin-transform-built-ins/src/builtins.js
vigneshshanmugam: Cool.. then will update

packages/babel-plugin-transform-built-ins/README.md
vigneshshanmugam: Alright

vigneshshanmugam:
@boopathi Addressed all the issues.

FezVrasta:
Is this already available in `0.0.11`?

vigneshshanmugam:
Nope.. Not released yet.

PR #416, opened 2017-02-10
don't try and minify props that contain unicode

### LINKED ISSUE #415: Syntax errors after minifying unicode
Trying to use [mathiasbynens/he](https://github.com/mathiasbynens/he) to encode html entities in the browser, but after adding the library, the app has started getting syntax errors in Safari and Firefox.

The offending character seems to be a unicode character, that appears in the source after minification.

Here's the result of running the corresponding snippet of code in [Babel's REPL with Babili enabled](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2015&targets=&browsers=&builtIns=false&code=let%20a%20%3D%20%7B%20'%5Cu2118'%3A'wp'%20%7D&experimental=false&loose=false&spec=false&playground=false):

```js
let a = { '\u2118':'wp' };
// becomes
'use strict';var a={℘:'wp'};
```

Visiting that link in Firefox 50.1.0 results in an "Illegal character error" in the REPL. Seeing as it works fine without Babili, I'm guessing this is a bug, rather than a deliberate constraint.

Pretty sure it happens in [babel-plugin-transform-property-literals](https://github.com/babel/babili/tree/master/packages/babel-plugin-transform-property-literals) as it turns the string into a literal key.

danprince:
Firefox 50.1.0 (one behind latest stable) and Safari 10.0.2 (latest stable) are unable to parse object literal properties that contain unicode characters which are not valid ES5 identifiers.

The minifier saves bytes by converting objects such as `{ '\u2118':'wp' }` into `{℘:'wp'}`, which only works when the key is a valid identifier.

The fix involves bypassing any properties with string representation containing characters outside of the `a-zA-Z0-9_$` range.

Closes #415.

### DIFF
--- a/packages/babel-plugin-transform-property-literals/src/index.js
+++ b/packages/babel-plugin-transform-property-literals/src/index.js
@@ -12,6 +12,11 @@ module.exports = function({ types: t }) {
             return;
           }
 
+          const containsUnicode = /[^a-z0-9$_]/i.test(key.value);
+          if (containsUnicode) {
+            return;
+          }
+
           if (key.value.match(/^\d+$/)) {
             node.key = t.numericLiteral(parseInt(node.key.value, 10));
             node.computed = false;
--- a/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
+++ b/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
@@ -50,4 +50,13 @@ describe("transform-property-literals-plugin", () => {
     `);
     expect(transform(source)).toBe(source);
   });
+
+  it("should not transform property keys that contain unicode", () => {
+    const source = unpad(`
+      ({
+        "\u2118": "wp"
+      });
+    `);
+    expect(transform(source)).toBe(source);
+  });
 });


mathiasbynens:
> Not all browsers can parse unicode in raw property names

This description is kind of vague. Which browsers? 

Anyway, here’s the underlying cause: https://mothereff.in/js-variables#%5Cu2118

> That’s a valid identifier according to ECMAScript 6 / Unicode 8.0.0.
> 
> However, it is not a valid identifier as per ES5. 

For example, `ლ` probably works in the browsers you were referring to, since it was already a valid identifier in ES5.

See https://mathiasbynens.be/notes/javascript-identifiers and https://mathiasbynens.be/notes/javascript-identifiers-es6 for background.

Can we amend the description accordingly?

We could narrow down the regex from `/[^a-z0-9$_]/i` to a regex that matches post-ES5 identifier chars only (when ES5 output is desired).

mathiasbynens:
> We could narrow down the `containsUnicode` regex from `/[^a-z0-9$_]/i` to a regex that matches post-ES5 identifier chars only (when ES5 output is desired).

This script generates a regex that matches ES2015 `IdentifierPart` except for ES5 `IdentifierPart`:

```js
// Note: run `npm install regenerate unicode-9.0.0` first!
// Based on https://gist.github.com/mathiasbynens/6334847.

const fs = require('fs');
const regenerate = require('regenerate');

// Which Unicode version should be used?
const version = '9.0.0';

// Set up a shorthand function to import Unicode data.
const get = function(what) {
	return require(`unicode-${ version }/${ what }/code-points.js`);
};

// Get the Unicode categories needed to construct the ES5 regex.
const Lu = get('General_Category/Uppercase_Letter');
const Ll = get('General_Category/Lowercase_Letter');
const Lt = get('General_Category/Titlecase_Letter');
const Lm = get('General_Category/Modifier_Letter');
const Lo = get('General_Category/Other_Letter');
const Nl = get('General_Category/Letter_Number');
const Mn = get('General_Category/Nonspacing_Mark');
const Mc = get('General_Category/Spacing_Mark');
const Nd = get('General_Category/Decimal_Number');
const Pc = get('General_Category/Connector_Punctuation');

// Get the Unicode properties needed to construct the ES6 regex.
const ID_Start = get('Binary_Property/ID_Start');
const ID_Continue = get('Binary_Property/ID_Continue');
const Other_ID_Start = get('Binary_Property/Other_ID_Start');

// https://mathiasbynens.be/notes/javascript-identifiers#valid-identifier-names
const es5IdentifierStart = regenerate('$', '_')
	.add(Lu, Ll, Lt, Lm, Lo, Nl)
	.removeRange(0x010000, 0x10FFFF); // Remove astral symbols.
const es5IdentifierPart = es5IdentifierStart.clone()
	.add('\u200C', '\u200D', Mn, Mc, Nd, Pc)
	.removeRange(0x010000, 0x10FFFF); // Remove astral symbols.

// http://ecma-international.org/ecma-262/6.0/#sec-identifier-names-static-semantics-early-errors
// http://unicode.org/reports/tr31/#Default_Identifier_Syntax
// https://bugs.ecmascript.org/show_bug.cgi?id=2717#c0
const es6IdentifierStart = regenerate(ID_Start)
	// Note: this already includes `Other_ID_Start`. http://git.io/wRCAfQ
	.add(
		'$',
		'_'
	);
const es6IdentifierPart = regenerate(ID_Continue)
	// Note: `ID_Continue` already includes `Other_ID_Continue`. http://git.io/wRCAfQ
	.add(Other_ID_Start)
	.add(
		'$',
		'_',
		'\u200C',
		'\u200D'
	);

const diff = es6IdentifierPart.clone()
	.remove(es5IdentifierPart);
const pattern = diff.toString();

console.log(`/${ pattern }/`);
```

The output is:

```js
/[\xB7\u0387\u1369-\u1371\u19DA\u2118\u212E\u309B\u309C]|\uD800[\uDC00-\uDC0B\uDC0D-\uDC26\uDC28-\uDC3A\uDC3C\uDC3D\uDC3F-\uDC4D\uDC50-\uDC5D\uDC80-\uDCFA\uDD40-\uDD74\uDDFD\uDE80-\uDE9C\uDEA0-\uDED0\uDEE0\uDF00-\uDF1F\uDF30-\uDF4A\uDF50-\uDF7A\uDF80-\uDF9D\uDFA0-\uDFC3\uDFC8-\uDFCF\uDFD1-\uDFD5]|\uD801[\uDC00-\uDC9D\uDCA0-\uDCA9\uDCB0-\uDCD3\uDCD8-\uDCFB\uDD00-\uDD27\uDD30-\uDD63\uDE00-\uDF36\uDF40-\uDF55\uDF60-\uDF67]|\uD802[\uDC00-\uDC05\uDC08\uDC0A-\uDC35\uDC37\uDC38\uDC3C\uDC3F-\uDC55\uDC60-\uDC76\uDC80-\uDC9E\uDCE0-\uDCF2\uDCF4\uDCF5\uDD00-\uDD15\uDD20-\uDD39\uDD80-\uDDB7\uDDBE\uDDBF\uDE00-\uDE03\uDE05\uDE06\uDE0C-\uDE13\uDE15-\uDE17\uDE19-\uDE33\uDE38-\uDE3A\uDE3F\uDE60-\uDE7C\uDE80-\uDE9C\uDEC0-\uDEC7\uDEC9-\uDEE6\uDF00-\uDF35\uDF40-\uDF55\uDF60-\uDF72\uDF80-\uDF91]|\uD803[\uDC00-\uDC48\uDC80-\uDCB2\uDCC0-\uDCF2]|\uD804[\uDC00-\uDC46\uDC66-\uDC6F\uDC7F-\uDCBA\uDCD0-\uDCE8\uDCF0-\uDCF9\uDD00-\uDD34\uDD36-\uDD3F\uDD50-\uDD73\uDD76\uDD80-\uDDC4\uDDCA-\uDDCC\uDDD0-\uDDDA\uDDDC\uDE00-\uDE11\uDE13-\uDE37\uDE3E\uDE80-\uDE86\uDE88\uDE8A-\uDE8D\uDE8F-\uDE9D\uDE9F-\uDEA8\uDEB0-\uDEEA\uDEF0-\uDEF9\uDF00-\uDF03\uDF05-\uDF0C\uDF0F\uDF10\uDF13-\uDF28\uDF2A-\uDF30\uDF32\uDF33\uDF35-\uDF39\uDF3C-\uDF44\uDF47\uDF48\uDF4B-\uDF4D\uDF50\uDF57\uDF5D-\uDF63\uDF66-\uDF6C\uDF70-\uDF74]|\uD805[\uDC00-\uDC4A\uDC50-\uDC59\uDC80-\uDCC5\uDCC7\uDCD0-\uDCD9\uDD80-\uDDB5\uDDB8-\uDDC0\uDDD8-\uDDDD\uDE00-\uDE40\uDE44\uDE50-\uDE59\uDE80-\uDEB7\uDEC0-\uDEC9\uDF00-\uDF19\uDF1D-\uDF2B\uDF30-\uDF39]|\uD806[\uDCA0-\uDCE9\uDCFF\uDEC0-\uDEF8]|\uD807[\uDC00-\uDC08\uDC0A-\uDC36\uDC38-\uDC40\uDC50-\uDC59\uDC72-\uDC8F\uDC92-\uDCA7\uDCA9-\uDCB6]|\uD808[\uDC00-\uDF99]|\uD809[\uDC00-\uDC6E\uDC80-\uDD43]|[\uD80C\uD81C-\uD820\uD840-\uD868\uD86A-\uD86C\uD86F-\uD872][\uDC00-\uDFFF]|\uD80D[\uDC00-\uDC2E]|\uD811[\uDC00-\uDE46]|\uD81A[\uDC00-\uDE38\uDE40-\uDE5E\uDE60-\uDE69\uDED0-\uDEED\uDEF0-\uDEF4\uDF00-\uDF36\uDF40-\uDF43\uDF50-\uDF59\uDF63-\uDF77\uDF7D-\uDF8F]|\uD81B[\uDF00-\uDF44\uDF50-\uDF7E\uDF8F-\uDF9F\uDFE0]|\uD821[\uDC00-\uDFEC]|\uD822[\uDC00-\uDEF2]|\uD82C[\uDC00\uDC01]|\uD82F[\uDC00-\uDC6A\uDC70-\uDC7C\uDC80-\uDC88\uDC90-\uDC99\uDC9D\uDC9E]|\uD834[\uDD65-\uDD69\uDD6D-\uDD72\uDD7B-\uDD82\uDD85-\uDD8B\uDDAA-\uDDAD\uDE42-\uDE44]|\uD835[\uDC00-\uDC54\uDC56-\uDC9C\uDC9E\uDC9F\uDCA2\uDCA5\uDCA6\uDCA9-\uDCAC\uDCAE-\uDCB9\uDCBB\uDCBD-\uDCC3\uDCC5-\uDD05\uDD07-\uDD0A\uDD0D-\uDD14\uDD16-\uDD1C\uDD1E-\uDD39\uDD3B-\uDD3E\uDD40-\uDD44\uDD46\uDD4A-\uDD50\uDD52-\uDEA5\uDEA8-\uDEC0\uDEC2-\uDEDA\uDEDC-\uDEFA\uDEFC-\uDF14\uDF16-\uDF34\uDF36-\uDF4E\uDF50-\uDF6E\uDF70-\uDF88\uDF8A-\uDFA8\uDFAA-\uDFC2\uDFC4-\uDFCB\uDFCE-\uDFFF]|\uD836[\uDE00-\uDE36\uDE3B-\uDE6C\uDE75\uDE84\uDE9B-\uDE9F\uDEA1-\uDEAF]|\uD838[\uDC00-\uDC06\uDC08-\uDC18\uDC1B-\uDC21\uDC23\uDC24\uDC26-\uDC2A]|\uD83A[\uDC00-\uDCC4\uDCD0-\uDCD6\uDD00-\uDD4A\uDD50-\uDD59]|\uD83B[\uDE00-\uDE03\uDE05-\uDE1F\uDE21\uDE22\uDE24\uDE27\uDE29-\uDE32\uDE34-\uDE37\uDE39\uDE3B\uDE42\uDE47\uDE49\uDE4B\uDE4D-\uDE4F\uDE51\uDE52\uDE54\uDE57\uDE59\uDE5B\uDE5D\uDE5F\uDE61\uDE62\uDE64\uDE67-\uDE6A\uDE6C-\uDE72\uDE74-\uDE77\uDE79-\uDE7C\uDE7E\uDE80-\uDE89\uDE8B-\uDE9B\uDEA1-\uDEA3\uDEA5-\uDEA9\uDEAB-\uDEBB]|\uD869[\uDC00-\uDED6\uDF00-\uDFFF]|\uD86D[\uDC00-\uDF34\uDF40-\uDFFF]|\uD86E[\uDC00-\uDC1D\uDC20-\uDFFF]|\uD873[\uDC00-\uDEA1]|\uD87E[\uDC00-\uDE1D]|\uDB40[\uDD00-\uDDEF]/
```

danprince:
> This description is kind of vague. Which browsers?

Updated the description.

Do you think the edge case of non-ES6 identifier compliant unicode is big enough to warrant adding another that extra complexity to the package? Seeing as this obviously hasn't broken for anyone else yet, I'd guess that there just aren't many people trying to use unicode in their property names.

mathiasbynens:
> Do you think the edge case […] is big enough to warrant adding […]

That edge case is exactly what this issue is about. The current patch works, but sometimes results in larger output than necessary. The patch + the regular expression above would special-case _only_ the identifiers that need special-casing, resulting in optimal output size, which might be more desirable for a minifier. It’s not up to me to make that call though :)

mathiasbynens:
> Updated the description.

Sorry — I meant the description in the commit message

packages/babel-plugin-transform-property-literals/src/index.js
```diff
@@ -12,6 +12,11 @@ module.exports = function({ types: t }) {
             return;
           }
 
+          const containsUnicode = key.value.match(/[^a-z0-9$_]/i);
```
mathiasbynens: Since you don’t care about the actual matches, let’s use `test` instead of `match`.

danprince:
> That edge case is exactly what this issue is about. The current patch works, but sometimes results in larger output than necessary.

Sure, but the only times it _ever_ results in larger output is when an object property contains a unicode character that isn't a valid ES6 identifier. Doesn't seem like that's going to happen often enough to be worth the computational overhead of having to check every object property name through that crazy regular expression.

> Sorry — I meant the description in the commit message

Got it.

mathiasbynens:
> Doesn't seem like that's going to happen often enough to be worth the computational overhead of having to check every object property name through that crazy regular expression.

Regular expressions are heavily optimized in all JS engines — I doubt the computation overhead is significant, but haven’t benchmarked it yet. For a build tool such as a minifier IMHO it’s worth it to spend a couple of extra milliseconds to produce optimal output. You only build once, but the output is consumed (presumably) more than that.

patrick-steele-idem:
I also ran into the problem when using Babili to minify [he](https://github.com/mathiasbynens/he). My vote would be to go with with the solution that @danprince provided for now since it solves the problem and unicode property names are sufficiently rare. Maybe a follow up PR for the more precise fix if someone thought that was needed? In the meantime, I'll have to use a fork of this plugin which is unfortunate, but it would be great to see this PR or another fix merged in and a new version published. Thanks.

patrick-steele-idem:
I did some additional investigation and found that the `babel-plugin-minify-constant-folding` plugin will also introduce `Identifier` nodes for object keys even if the key name contains special unicode characters because `t.isValidateIdentifier()` doesn't restrict to safe identifiers (safe for currently the latest Safari v10.0.3, that is). The `babel-plugin-transform-property-literals` plugin only transforms string literal keys (not identifier keys). Therefore, even with this patch, `he` does not get minified in a way that can be loaded in the latest Safari.

I tried adding additional plugins/presets to run after babili to convert the unsafe object keys back to string literals, but for some reason babili plugins always ran last. Any ideas?

I decided to create another PR that only allows safe object keys to be used as identifiers: https://github.com/babel/babili/pull/445

Please consider merging that PR because the current babili does not always produce JS code that will load in all browsers.

timsuchanek:
Yes, would be great if the PR gets merged as this stops us from using babili (which we need in order to have webpack treeshaking with TypeScript)
Is there anything blocking the merge where help is needed?

PR #375, opened 2017-01-14
Account for removal of consequent inside ternary for no-console plugin. Fixes #374

DrewML:
Figured I'd take a whack at it.  The ideal minified form of this would probably be:
```js
// in
var foo = console.log ? console.log('lol') : someOtherFn;
// out
var foo = someOtherFn;
 ```

but I just went with what would fix the bug for now. Feedback welcome.

PR #421, opened 2017-02-12
Fix remove console.* statements

### LINKED ISSUE #374: SyntaxError after transform-remove-console
Original bug report from @brunolemos in https://github.com/babel/babel/issues/5106
---------

### Input Code
https://github.com/jayphelps/core-decorators.js/blob/master/src/time.js#L7

```js
var defaultConsole = {
  time: console.time ? console.time.bind(console) : function (label) {
    labels[label] = new Date();
  },
};
```

### Babel Configuration (.babelrc, package.json, cli command)
```json
{
  "presets": [
    "react-native",
    "react-native-stage-0/decorator-support"
  ],
  "plugins": [
    "transform-es2015-parameters",
    "transform-remove-console"
  ]
}
```

### Expected Behavior
Remove console.* without breaking the code

### Current Behavior
```shell
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1740):
SyntaxError: TransformError: /node_modules/core-decorators/lib/time.js: /node_modules/core-decorators/lib/time.js:
Property consequent of ConditionalExpression expected node to be of a type ["Expression"] but instead got null
(node:47674) DeprecationWarning: Unhandled promise rejections are deprecated. 
In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### Context
Added `transform-remove-console` on a React Native application, but it crashes.

### Your Environment
| software         | version
| ------------- | -------
| Babel              |  6.18
| node               | 7.3.0
| npm                | 4.0.5
| React Native  | 0.39.2
| Operating System | macOS 10.12.3

boopathi:
+ (Fix #374)
+ (Fix #375)

### DIFF
--- a/packages/babel-plugin-transform-remove-console/src/index.js
+++ b/packages/babel-plugin-transform-remove-console/src/index.js
@@ -24,11 +24,15 @@ module.exports = function({ types: t }) {
       MemberExpression: {
         exit(path) {
           if (isConsole(path) && !path.parentPath.isMemberExpression()) {
-            path.replaceWith(createNoop());
+            if (path.parentPath.isAssignmentExpression() && path.parentKey === "left") {
+              path.parentPath.get("right").replaceWith(createNoop());
+            } else {
+              path.replaceWith(createNoop());
+            }
           }
         }
       }
-    },
+    }
   };
 
   function isGlobalConsoleId(id) {
--- a/packages/babel-plugin-transform-remove-console/src/index.js
+++ b/packages/babel-plugin-transform-remove-console/src/index.js
@@ -23,8 +23,7 @@ module.exports = function({ types: t }) {
       },
       MemberExpression: {
         exit(path) {
-          if (!isConsole(path)) return;
-          if (!path.parentPath.isMemberExpression()) {
+          if (isConsole(path) && !path.parentPath.isMemberExpression()) {
             path.replaceWith(createNoop());
           }
         }
--- a/packages/babel-plugin-transform-remove-console/src/index.js
+++ b/packages/babel-plugin-transform-remove-console/src/index.js
@@ -33,7 +33,15 @@ module.exports = function({ types: t }) {
   };
 
   function isConsole(memberExpr) {
-    return memberExpr.get("object").isIdentifier({ name: "console" });
+    const object = memberExpr.get("object");
+    if (object.isIdentifier({ name: "console" })) return true;
+
+    const property = memberExpr.get("property");
+    return object.get("object").isIdentifier({ name: "console" })
+      && (
+        property.isIdentifier({ name: "call" })
+        || property.isIdentifier({ name: "apply" })
+      );
   }
 
   function isConsoleBind(memberExpr) {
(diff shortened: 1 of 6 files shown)


vigneshshanmugam:
👍

mikesherov:
Hi! I'm a first time contributor to babili, and had a few questions/comments about this pull request!

1. Considering babili is es6 aware, why the change from arrowFunctionExpressions to normal ones?
2. I think this PR misses console.log.call and console.log.apply.

Thanks!

hzoo:
Wow @mikesherov's here 😄

packages/babel-plugin-transform-remove-console/src/index.js
```diff
@@ -1,14 +1,53 @@
...
+          if (!isConsole(path)) return;
+          if (!path.parentPath.isMemberExpression()) {
+            path.replaceWith(createNoop());
+          }
+        }
+      }
     },
   };
+
+  function isConsole(memberExpr) {
```
mikesherov: Seems this needs to check for console.log.call and console.log.apply.

mikesherov: And the test needs to check both `a = console.log.call` and `console.log.call()`

boopathi:
@mikesherov  this will change it(not transforming function to arrow. But use arrow directly by checking if the target the user intended supports arrow) - https://github.com/babel/babili/issues/428.

mikesherov:
Got it.

boopathi:
Added `call` and `apply`.

packages/babel-plugin-transform-remove-console/src/index.js
```diff
@@ -1,14 +1,61 @@
...
+          }
+        } else if (isConsoleBind(callee)) {
+          // console.log.bind()
+          path.replaceWith(createNoop());
         }
       },
+      MemberExpression: {
+        exit(path) {
+          if (!isConsole(path)) return;
+          if (!path.parentPath.isMemberExpression()) {
```
boopathi: Yup!. Thanks.

packages/babel-plugin-transform-remove-console/src/index.js
```diff
@@ -1,14 +1,67 @@
...
+        } else if (isConsoleBind(callee)) {
+          // console.log.bind()
+          path.replaceWith(createNoop());
         }
       },
+      MemberExpression: {
+        exit(path) {
+          if (isConsole(path) && !path.parentPath.isMemberExpression()) {
+            path.replaceWith(createNoop());
+          }
```
jhen0409: It looks not work with `console.err = xx` ([React Native have similar usage](https://github.com/facebook/react-native/blob/master/Libraries/Core/ExceptionsManager.js#L103-L104)), maybe we can just `parentPath.remove()` if path key is `left` and parentPath is assignment expression?

boopathi: Interesting, So, I would assume it to be replaced like this - 

```js
// in
console.foo = foo;
// out
console.foo = function () {}; 
```

which looks like the right thing to do. because it is hard to detect all usages of `console.foo` in code 

```js
const x = "foo";
const y = console;
y[x]();
```

and will lead to runtime error.

PR #446, opened 2017-02-28
Minor optimization for handling blacklisted names

hzlmn:
Currently  if `array` passed as `blacklist` value  in options, `isBlacklist` method always going through the list of props.

My version instead iterate over array once and create map structure with const value accessing time.

### DIFF
--- a/packages/babel-plugin-minify-mangle-names/src/index.js
+++ b/packages/babel-plugin-minify-mangle-names/src/index.js
@@ -21,7 +21,7 @@ module.exports = ({ types: t, traverse }) => {
     } = {}) {
       this.charset = charset;
       this.program = program;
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-mangle-names/src/index.js
```diff
@@ -311,6 +308,19 @@ class Charset {
   }
 }
 
+// convert value to Map object
+function toMap(value) {
```
boopathi: Since JS has a [Map](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map) let's call this `toObject`

hzlmn: @boopathi Updated

Issue #425, opened 2017-02-17
minify-builtin aliases are not mangled by default


boopathi:
requires passing `topLevel` option to mangler. There should be a way to mangle these names that we generate by default without the topLevel option. 

/cc @vigneshshanmugam 

```js
function foo() {
  return Math.floor(bar);
}
```

current output: 

```js
var _Mathfloor = Math.floor;
function foo() {
  return _Mathfloor(bar);
}
```

should be

```js
var a = Math.floor;
function foo() { // foo is NOT mangled
  return a(bar);
}
```

PR #472, opened 2017-03-09
fix builtins plugins from leaking vars

### LINKED ISSUE #461: minify-builtins leaks vars
minify-builtins leaks the variables it produces. It needs to look and see if the rest of the code is in an IIFE and put its vars inside if so. And being outside the IIFE, the vars aren't mangled (#425) so that it doesn't decrease the file size and probably not even the gzipped size!

vigneshshanmugam:
fix #461 
+ The implementation decides where to insert the created identifier based on the input code. 

Even after fixing this, the mangler doesn't rename the identifier properly. Related - https://github.com/babel/babili/issues/425

### DIFF
--- a/packages/babel-plugin-minify-builtins/src/index.js
+++ b/packages/babel-plugin-minify-builtins/src/index.js
@@ -84,7 +84,7 @@ module.exports = function({ types: t }) {
             path.replaceWith(uniqueIdentifier);
           }
           // hoist the created var to the top of the block/program
-          const path = getPath(paths[0], paths[paths.length - 1]);
+          const path = getLeastCommonFunctionPath(paths[0], paths[paths.length - 1]);
           path.unshiftContainer("body", newNode);
         }
       }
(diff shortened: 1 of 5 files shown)


curiousdannii:
If I'm reading the tests correctly this still leaks some vars. Unless it can detect it's in module mode it would be better not to try to convert top level builtins.

packages/babel-plugin-minify-builtins/src/index.js
```diff
@@ -123,6 +124,26 @@ module.exports = function({ types: t }) {
...
 
+// decides the best path to insert the newly created identifier
+function getPath(program, firstPath, lastPath) {
+  let resultPath = program;
+  // check if occurence of member exp ex - Max.max() is before the
+  // block statement container
+  const { start: nodeStart } = firstPath.parent;
+  const { end: nodeEnd } = lastPath.parent;
+
+  program.traverse({
```
boopathi: the same can be achieved by `getFunctionParent` recursively till `program`. `program.traverse` inside a loop will be costly.

vigneshshanmugam:
[Update]

Running mangler on exit fixes the #425 issue.

vigneshshanmugam:
Works now for functions at same level.

PR #487, opened 2017-04-03
[WIP] Upgrade to Babel 7

boopathi:
Babel 7 API upgrade

Current Babel version: `7.0.0-alpha.6`

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,12 +36,11 @@
     "coverageDirectory": "./coverage/"
   },
   "devDependencies": {
-    "babel-core": "^6.24.0",
+    "babel-core": "7.0.0-alpha.6",
     "babel-eslint": "^7.2.1",
     "babel-jest": "^19.0.0",
-    "babel-plugin-transform-es2015-block-scoping": "^6.23.0",
     "babel-preset-env": "^1.2.2",
-    "babel-traverse": "^6.23.1",
+    "babel-traverse": "7.0.0-alpha.3",
     "bytes": "^2.5.0",
     "chalk": "^1.1.3",
     "cli-table": "^0.3.1",
--- a/packages/babel-helper-mark-eval-scopes/__tests__/helper-mark-eval-scopes-test.js
+++ b/packages/babel-helper-mark-eval-scopes/__tests__/helper-mark-eval-scopes-test.js
@@ -10,7 +10,7 @@ function getPath(source) {
     babelrc: false,
     plugins: [
       function({ traverse }) {
-        traverse.clearCache();
+        traverse.cache.clear();
         return {
           visitor: {
             Program(programPath) {
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -842,7 +842,7 @@ module.exports = ({ types: t, traverse }) => {
             } = {}
           } = {}
         ) {
-          traverse.clearCache();
+          traverse.cache.clear();
           path.scope.crawl();
 
           markEvalScopes(path);
--- a/packages/babel-plugin-minify-mangle-names/__tests__/mangle-names-test.js
+++ b/packages/babel-plugin-minify-mangle-names/__tests__/mangle-names-test.js
@@ -682,7 +682,7 @@ describe("mangle-names", () => {
       code: false
     });
 
-    traverse.clearCache();
+    traverse.cache.clear();
 
     const actual = babel.transformFromAst(first.ast, null, {
       plugins: [require("../src/index")]
@@ -734,7 +734,7 @@ describe("mangle-names", () => {
       code: false
     });
 
-    traverse.clearCache();
+    traverse.cache.clear();
 
     const actual = babel.transformFromAst(first.ast, null, {
       plugins: [require("../src/index")]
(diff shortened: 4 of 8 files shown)


hzoo:
Hey @boopathi! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node v4 - linux](https://travis-ci.org/babel/babili/jobs/218115093)

package.json
```diff
@@ -36,12 +36,11 @@
     "coverageDirectory": "./coverage/"
   },
   "devDependencies": {
-    "babel-core": "^6.24.0",
+    "babel-core": "7.0.0-alpha.6",
     "babel-eslint": "^7.2.1",
     "babel-jest": "^19.0.0",
-    "babel-plugin-transform-es2015-block-scoping": "^6.23.0",
     "babel-preset-env": "^1.2.2",
```
hzoo: there's also a v2.0.0-alpha.3 for this if we want to use

diervo:
Any blocker to push this, apart from probably increment minor version?

hzoo:
If we release as a minor or anything it also means that users using babili as a preset will fail unless they use babel-core 7.0 alpha as well

diervo:
Why don't generate a next version instead of latest like all of the 7beta packages?

diervo:
@boopathi @hzoo This is preventing me for upgrading all our build process to babel7, and I would argue that many other developers too (to reuse the same AST tree)

I will be happy to maintain a branch in sync with upstream changes if that's required.
Is there any other blockers? 
Otherwise I would love to have a resolution on this or a timeframe/plan at least.

kelset:
Any ETA for this PR to be merged?

boopathi:
Waiting to find a way to continue testing packages for `babel 6` too. Along with babel-7.

vigneshshanmugam:
Trying out babel-7.0.0-beta.0 version.

package.json
```diff
@@ -70,5 +71,8 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "resolutions": {
```
vigneshshanmugam: This is require because

1. jest-cli uses babel-core: 6 version which is hoisted to the node_modules directory by using yarn workspaces which in turn hoists babel-register: 6
2. running babel-register with 6 version would break and report this following error which is deprecated in babel 7
```js
TypeError: /Users/vshanmugam/open-source/babili/gulpfile.babel.js: path.arrowFunctionToExpression is not a function
    at PluginPass.ArrowFunctionExpression (/Users/vshanmugam/open-source/babili/node_modules/babel-plugin-transform-es2015-arrow-functions/lib/index.js:10:14)

```

remcohaszing:
Some updates:

Official `babel-*` packages have been renamed to `@babel/*`.

`babel-minify@0.2.0` seems to work after upgrading babel to the namespaced beta packages based on a `package-lock.json` file, but it seems to break when the lock file is removed and a new one is created. (I encountered this in [this](https://github.com/remcohaszing/spellcard-creator) project).

As can be seen in the `babel-core` [options](https://github.com/babel/babel/tree/master/packages/babel-core#options), Babel now handles whitespaces internally based on the `compact` option. This causes most test failures. The output of the tests can be modified by specifying the `compact` parameter in [test-transform.js](https://github.com/babel/minify/blob/babel-7/utils/test-transform.js#L25). Either way, many tests will need some manual whitespace modifications.

- Should the babel-7 branch just focus on Babel 7 only and drop support for Babel 6 because of the rename?
- Should babel-minify use namespaced packages as well?
- Is compact mode (more real) or non compact mode (more readable for tests) preferred?

I may want to try and fix this port, or at least partially, if these questions get answered.

lukescott:
Any updates on this? Went to add babel/minify to an existing project that was updated to babel 7 and discovered it wasn't compatible.

To the first two questions, I would assume yes. Since minify is part of the babel org, I would assume all packages should live under that umbrella. And since the package names would change, it would be easy to drop support for babel 6 and let people use the older package names for babel 6.

packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
```diff
@@ -241,3 +241,11 @@ module.exports = class ScopeTracker {
     bindings.delete(oldName);
   }
 };
+
+/**
+ * Babel-7 returns null if there is no function parent
+ * and uses getProgramParent to get Program
+ */
+function getFunctionParent(path) {
```
vigneshshanmugam: does it make sense to make it to one of the helpers?

diervo:
@boopathi thanks a bunch for rebasing the work, will continue our testing!
Is there anyway we can help get a stable beta branch out soon?

We will be using babili on production shortly, so our team at Salesforce can help with some bugs and contributions whenever possible.

hzoo:
Yeah although I haven't been involved in this particular project for a while I think it should land on master and then make another branch for the old version and start doing releases? Assuming the tests pass already

diervo:
That would  be awesome!
Yes all tests are passing, I created my own fork where I'm testing the integration with another project that depends on babel7.

I will report back more issues soon ;)

vigneshshanmugam:
> Yeah although I haven't been involved in this particular project for a while I think it should land on master and then make another branch for the old version and start doing releases? This is what we did with v7; we spent a long time making a 7.x branch but it was easier to just develop on master and create a 6.x branch for old code.

I would say so, Makes it better to do a patch release and backport it to 6 when needed while we work on new features in master.

boopathi:
Master now uses babel-7 🎉

`npm install babel-minify@canary` should now install babel-7 version (auto published from master). 

babel-6 is in `0.3.x` branch.

PR #489, opened 2017-04-03
Create benchmark_cache dir if it doesn't exist

dirtybit:
Create benchmark_cache directory if it doesn't exist before downloading the default packages.


boopathi:
Thanks!

PR #490, opened 2017-04-03
Remove jsesc

### LINKED ISSUE #413: Invalid string concatenation breaks Lodash _.template method
Minifing the Lodash library using Babili 0.0.11 breaks the `_.template` method.

The string concatenation in that method (https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L14748) is minified to contain extraneous escape characters.

Input:
```
source = 'function(' + (variable || 'obj') + ') {\n' +
        (variable
          ? ''
          : 'obj || (obj = {});\n'
        ) +
        "var __t, __p = ''" +
        (isEscaping
           ? ', __e = _.escape'
           : ''
        ) +
        (isEvaluating
          ? ', __j = Array.prototype.join;\n' +
            "function print() { __p += __j.call(arguments, '') }\n"
          : ';\n'
        ) +
        source +
        'return __p\n}';
```

Output:
`source='function('+(variable||'obj')+') {\n'+(variable?'':'obj || (obj = {});\n')+'var __t, __p = \'\''+(isEscaping?', __e = _.escape':'')+(isEvaluating?', __j = Array.prototype.join;\\nfunction print() { __p += __j.call(arguments, \\\'\\\') }\\n':';\n')+source+'return __p\n}';`

Notice extra escape in `\\n` which results in `\n` in function body after `Array.prototype.join`:
`... Array.prototype.join; \nfunction print() {...`

That causes `SyntaxError: Invalid or unexpected token` when function is executed.

### LINKED ISSUE #440: unnecessary escaping in string concatinaion
Hey, 
For some reason I can't explain babili(?) escapes the backslash in the following string:
input:
```js
var x =  '\'cool\'' + 'test'
```
babili output:
```js
var x='\\\'cool\\\'test';
```
it only happens when i concat strings, it works perfectly when i input:
```js
var x =  '\'cool\'test'
```

I use the babili cli without any presets

p.s the [babel repl](https://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=&targets=&browsers=&builtIns=false&code=var%20x%20%3D%20%20'%5C'cool%5C''%20%2B%20'test') doesn't show the problem (probably browser html escaping)

boopathi:
+ Fix #440, fix #413
+ Related #384, #382

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
@@ -3,9 +3,9 @@ jest.autoMockOff();
 const babel = require("babel-core");
 const unpad = require("../../../utils/unpad");
 
-function transform(code, opts = {}) {
+function transform(code) {
   return babel.transform(code, {
-    plugins: [[require("../src/index"), opts]]
+    plugins: [require("../src/index")]
   }).code;
 }
 
@@ -71,7 +71,7 @@ describe("constant-folding-plugin", () => {
     expect(transform(source)).toBe(source);
   });
 
-  it("should handle script escape", () => {
+  xit("should handle script escape", () => {
     const source = unpad(
       `
       "</" + "script"
@@ -83,10 +83,10 @@ describe("constant-folding-plugin", () => {
       "<\\\\/script";
     `
     );
-    expect(transform(source, { isScriptContext: true })).toBe(expected);
+    expect(transform(source)).toBe(expected);
   });
 
-  it("should handle style escape", () => {
+  xit("should handle style escape", () => {
     const source = unpad(
       `
       "</" + "style"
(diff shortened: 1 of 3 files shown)


goto-bus-stop:
Hmm, the problem in #413/#440 isn't `isScriptContext`, but all the other things that jsesc does (like escaping backslashes). `isScriptContext` was the solution to the earlier problem in #382, where `</script>` could end up in the output inlined in an HTML `<script>` tag.

A solution entirely in the constant folding plugin would be to manually replace `</script>` → `<\/script>` and `<!--` to `\x3C!--`, which is what the `isScriptContext` option does

boopathi:
Yes. The problem is running jsesc twice. Babel generator already passes all [String Literals via jsesc](https://github.com/babel/babel/blob/8a82cc060ae0ab46bf52e05e592de770bd246f6f/packages/babel-generator/src/generators/types.js#L137). I'll add a PR to babel to pass isScriptContext to the generator- but probably will be added to the current track - 7.0. And remove jsesc from babili and this option after that.

goto-bus-stop:
Yes but this would still run jsesc twice, wouldn't it?

boopathi:
Will remove jsesc after https://github.com/babel/babel/pull/5581 .

boopathi:
Merging this and reopening #382 . #382 will be fixed via an option in babel.

mathiasbynens:
For context, see https://github.com/babel/babili/issues/440#issuecomment-290413795. This patch
 is a part of that solution. 👍

boopathi:
Yes !! https://github.com/babel/babel/pull/5581 ..

PR #506, opened 2017-04-18
avoid transforming side effecty member expression checks[fix #469]

### LINKED ISSUE #469: babel-plugin-minify-dead-code-elimination: produces wrong code
version 0.1.4

from Babel REPL:
```javascript
 function subList(list=[], template) {

      var olElm = template.element;

      if (olElm.tagName !== 'OL') {
        olElm = document.createElement('ol');
      } else {
        var reinsert = foo => list.push(olElm);
        var hash = {};
      }

      if (hash) {
        reinsert();
      }

      return olElm;
    }
```

produces:

```javascript
function subList(a=[],b){
var c=b.element;
if('OL'!==c.tagName)
  c=document.createElement('ol');
else
  var d=()=>a.push(c);
return d(),c
}
```

`hash` is eliminated when it shouldn't be. This is a cut down version of the code that had the problem which is why it doesn't make much sense.

vigneshshanmugam:
+ this transform happens on simplify as well.. So the issue itself is not resolved. 

fix #469

Just keeping it here so that we can remove it from one place and apply this fix.

### DIFF
--- a/packages/babel-plugin-minify-guarded-expressions/__tests__/guarded-expressions-test.js
+++ b/packages/babel-plugin-minify-guarded-expressions/__tests__/guarded-expressions-test.js
@@ -178,4 +178,16 @@ describe("guarded-expressions-plugin", () => {
     );
     expect(transform(source)).toBe(source);
   });
+
+  it("should not minify side effecty member expression checks", () => {
+    let source = unpad(
+      `
+      {
+        var a = {};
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
@boopathi Does this still happen? Can we merge or close this since we need this for 1.0 milestone

PR #517, opened 2017-05-04
Fix crashing in simplify plugin. Deopt instead

### LINKED ISSUE #455: Crash minifying in babili 0.0.12
Minifying this in babili 0.0.12:

```
function foo(param)
{
	if (param === null)
		return;
	
	let thingA = param.a;
	let thingB = param.b;
	
	if (!thingA && !thingB)
		return;
	
	let thingC = param.c;
}
```

crashes with "TypeError: unknown: Cannot read property 'scope' of null".

That's a minimal repro, if you delete any lines from the function foo it doesn't reproduce any more.

boopathi:
+ Fix #455

### DIFF
--- a/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
+++ b/packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
@@ -3152,4 +3152,32 @@ describe("simplify-plugin", () => {
     );
     expect(transform(source)).toBe(expected);
   });
+
+  it("should fix#455 and deopt when scope tree is not updated", () => {
+    const source = unpad(
+      `
+      function foo(param) {
+        if (param === null) return;
+        let thingA = param.a;
+        let thingB = param.b;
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-simplify/__tests__/simplify-test.js
```diff
@@ -3152,4 +3152,32 @@ describe("simplify-plugin", () => {
...
+    );
+    const expected = unpad(
+      `
+      function foo(param) {
+        if (param !== null) {
+            let thingA = param.a;
+            let thingB = param.b;
+            if (thingA || thingB) {
+                let thingC = param.c;
+              }
```
vigneshshanmugam: indentation is screwed up here.

boopathi: yeah. Has happened before for if_return.

PR #523, opened 2017-05-05
Fix collapse with circular reference

jhen0409:
This PR fix the bug for:

```js
var foo = {};
foo.bar = foo;

var foo2 = [];
foo2[2] = foo2;

var foo3 = new Set([])
foo3.add(foo3)
```

Will convert to:

```js
var foo = {bar:foo};

var foo2 = [,,foo2];

var foo3 = new Set([foo3])
```

It will get `undefined` for `var`, a ReferenceError for `let` and `const`.

I catched the problem with [`prop-types`](https://github.com/reactjs/prop-types/blob/master/factoryWithThrowingShims.js#L51) module.

### DIFF
--- a/packages/babel-plugin-transform-inline-consecutive-adds/src/index.js
+++ b/packages/babel-plugin-transform-inline-consecutive-adds/src/index.js
@@ -142,7 +142,7 @@ function getContiguousStatementsAndExpressions(
 
 function getReferenceChecker(references) {
   // returns a function s.t. given an expr, returns true iff expr is an ancestor of a reference
-  return expr => references.some(r => r.isDescendant(expr));
+  return expr => references.some(r => r === expr || r.isDescendant(expr));
 }
 
 function tryUseCollapser(t, collapser, varDecl, topLevel, checkReference) {
--- a/packages/babel-plugin-transform-inline-consecutive-adds/src/set-collapser.js
+++ b/packages/babel-plugin-transform-inline-consecutive-adds/src/set-collapser.js
@@ -44,7 +44,7 @@ class SetCollapser extends Collapser {
       if (args.length !== 1) {
         return false;
       }
-      if (checkReference(args)) {
+      if (checkReference(args[0])) {
         return false;
       }
       return true;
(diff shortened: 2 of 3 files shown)


packages/babel-plugin-transform-inline-consecutive-adds/src/object-collapser.js
```diff
@@ -39,6 +39,13 @@ class ObjectCollapser extends Collapser {
       if (checkReference(right)) {
         return false;
       }
+      if (
+        obj.isIdentifier() &&
+        right.isIdentifier() &&
+        obj.node.name === right.node.name
+      ) {
```
boopathi: Good catch. But, this check should go into the `checkReference` call which is used above so it applies the fix for all collapsers.

jhen0409: You're right! I'll do it to `checkReference`.

packages/babel-plugin-transform-inline-consecutive-adds/src/set-collapser.js
```diff
@@ -44,7 +44,7 @@ class SetCollapser extends Collapser {
       if (args.length !== 1) {
         return false;
       }
-      if (checkReference(args)) {
+      if (checkReference(args[0])) {
```
jhen0409: It should not expect an array (as Line 44), so I change to `args[0]`  made `Set` works with this PR.

packages/babel-plugin-transform-inline-consecutive-adds/src/index.js
```diff
@@ -142,7 +142,7 @@ function getContiguousStatementsAndExpressions(
 
 function getReferenceChecker(references) {
   // returns a function s.t. given an expr, returns true iff expr is an ancestor of a reference
-  return expr => references.some(r => r.isDescendant(expr));
+  return expr => references.some(r => r.isDescendant(expr) || r === expr);
```
boopathi: Isn't it better to place `r === expr` first for early exit ?

PR #527, opened 2017-05-10
Add butternut to benchmarks

boopathi:
https://github.com/Rich-Harris/butternut

/cc @Rich-Harris


README.md
```diff
@@ -189,23 +189,25 @@ Input Size (gzip): 29.82kB
 
 | minifier              | output raw  | raw win | gzip output | gzip win | parse time (ms) | minify time (ms) |
 | --------------------- | ----------- | ------- | ----------- | -------- | --------------- | ---------------- |
-| **babili**            | 36.13kB     | 71%     | 12.6kB      | 58%      | 3.14            | 1995.80          |
-| **uglify**            | **35.73kB** | **71%** | **11.96kB** | **60%**  | **2.28**        | **770.76**       |
-| **closureCompiler**   | 34.77kB     | 72%     | 11.96kB     | 60%      | 2.62            | 3940.01          |
-| **closureCompilerJs** | 65.41kB     | 48%     | 15.83kB     | 47%      | 2.96            | 1049.21          |
+| **babili**            | 35.86kB     | 71%     | 12.47kB     | 58%      | 8.14            | 1418.56          |
+| **uglify**            | 35.74kB     | 71%     | 11.97kB     | 60%      | 2.60            | 676.57           |
```
vigneshshanmugam: lets add uglify-es as well?

scripts/benchmark.js
```diff
@@ -128,6 +130,9 @@ class Benchmark {
     const out = compile(flags);
     return out.compiledCode;
   }
+  butternut(code) {
+    return butternut.squash(code, { check: true }).code;
```
Rich-Harris: Using the `check: true` option is a good idea to check that Butternut is producing valid code, but it *shouldn't* be used for benchmarking as it causes the output to be parsed again.

Also, Butternut generates a sourcemap by default. Since Babili and Uglify don't, we should probably disable sourcemaps for Butternut too to remove that from the equation:

```diff
-return butternut.squash(code, { check: true }).code;
+return butternut.squash(code, { sourceMap: false }).code;
```

boopathi: 👍

darahak:
What is the rule for using bold text in the benchmark tables? Highlight best results?
It seems inconsistent.

PR #528, opened 2017-05-10
Add gulp-babili dependencies.

### LINKED ISSUE #525: gulp-babili does not list all of its dependencies
`index.js` of gul-babili requires `through2` and `gulp-util`, but does not list them in package.json. Is this intended or a mistake?

andrewstart:
Fixes #525


boopathi:
Thanks!

PR #531, opened 2017-05-10
Add whitelist/blacklist to transform-inline-environment-variables

### LINKED ISSUE #513: babel-plugin-transform-inline-environment-variables - Blacklist/whitelist?
I've been using `babel-plugin-node-env-inline` for a while when compiling code that is to be used in Node. I'm getting to the point where I would like to inline other environment variables for dead code elimination while not inlining others for use with `dotenv`. Would it make sense to have an optional blacklist/whitelist to specify which environment variables should be inlined or not?

JakeDluhy:
Fixes #513

### DIFF
--- a/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
+++ b/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
@@ -1,14 +1,16 @@
 jest.autoMockOff();
 
-const thePlugin = require("../../../utils/test-transform")(
-  require("../src/index")
-);
+const plugin = require("../src/index")
+const thePlugin = require("../../../utils/test-transform")(plugin);
 
 describe("inline-env-plugin", () => {
   let prev;
   beforeAll(() => {
     prev = process.env.NODE_ENV;
     process.env.NODE_ENV = "development";
+  });
+  afterAll(() => {
+    process.env.NODE_ENV = prev;
   });
 
   thePlugin(
@@ -21,7 +23,21 @@ describe("inline-env-plugin", () => {
   `
   );
 
-  afterAll(() => {
-    process.env.NODE_ENV = prev;
-  });
+  thePlugin(
+    "should only include whitelisted variables if include option is specified",
+    "process.env.NODE_ENV",
+    "process.env.NODE_ENV;",
+    {
+      plugins: [[plugin, { include: ["IS_ELECTRON"] }]]
+    }
+  );
+
+  thePlugin(
+    "should not include blacklisted variables if exclude option is specified",
+    "process.env.NODE_ENV",
+    "process.env.NODE_ENV;",
+    {
+      plugins: [[plugin, { exclude: ["NODE_ENV"] }]]
+    }
+  );
 });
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
```diff
@@ -20,4 +20,28 @@ describe("inline-env-plugin", () => {
...
+  });
+
+  it("should not include blacklisted variables if exclude option is specified", () => {
+    const prev = process.env.NODE_ENV;
+    process.env.NODE_ENV = "development";
+    const exclude = ["NODE_ENV"];
+
+    const source = "process.env.NODE_ENV";
+    const expected = "process.env.NODE_ENV;";
+    expect(transform(source, { exclude })).toBe(expected);
```
vigneshshanmugam: why NODE_ENV is present in output even if its excluded?

JakeDluhy: My intention was that `exclude` means environment variables to exclude from being transformed, so it would be left as is.

packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
```diff
@@ -20,4 +20,28 @@ describe("inline-env-plugin", () => {
...
   });
+
+  it("should only include whitelisted variables if include option is specified", () => {
+    const prev = process.env.NODE_ENV;
+    process.env.NODE_ENV = "development";
+    const include = ["IS_ELECTRON"];
+
+    const source = "process.env.NODE_ENV";
+    const expected = "process.env.NODE_ENV;";
+    expect(transform(source, { include })).toBe(expected);
```
vigneshshanmugam: the expected value should be "development" here. Also can you please add process.env.IS_ELECTRON to the source in the test.

JakeDluhy: The test is testing that only variables that are specified in the `include` param get transformed. Since `IS_ELECTRON` is the only variable specified in `include`, `process.env.NODE_ENV` should _not_ be transformed.

I'm not exactly sure what you mean by adding `process.env.IS_ELECTRON` to the source, but I'm interpreting it as make sure that it gets transformed?

JakeDluhy: Can I please get further feedback that I'm correctly interpreting the comment?

JakeDluhy: @vigneshshanmugam ping

JakeDluhy:
Rebased on top of the latest master. Would still love to get this in.

christophehurpeau:
@boopathi hello, any reason why this is blocked ? Thanks !

j-f1:
@boopathi Closing + Reopening might fix that.

christophehurpeau:
@boopathi PR https://github.com/babel/minify/pull/672 but with options renamed `whitelist` and `blacklist` and tests passing :) Which option name are better ? I can change them back.

boopathi:
@JakeDluhy There is another PR (#672) for the same sent by @christophehurpeau . You sent the PR first. So I'll wait for you to fix the build. Do you want to fix this PR or is it fine if we merge the new PR #672 ?

JakeDluhy:
@boopathi thanks! I fixed the PR so it should be good to go

kochis:
@boopathi is it possible to push a release to just `babel-plugin-transform-inline-environment-variables` with these changes?

PR #533, opened 2017-05-11
handle deopt case in builtins properly

### LINKED ISSUE #458: babilified code fails at runtime on SystemJS
I attach the original and minified code (it is a SystemJS bundle produced via jspm bundle, minified with babili 6.22.2). the original works correctly when loaded with SystemJS and executing

```javascript
SystemJS.import('icojs')
```

but the minified code throws an error at runtime (`Error: Te is not a constructor`). I have minified other SystemJS bundles without issue before, so this happens to be an exceptional case.

[original.txt](https://github.com/babel/babili/files/819386/original.txt)

[minified.txt](https://github.com/babel/babili/files/819387/minified.txt)

### LINKED ISSUE #529: Hoisted or aliased variable issue with webpack
My build with webpack and babili produced that kind of code

```js
var _Mathround = Math.round;
webpackJsonp([11], {
    1022: function(a, b, c) {
        'use strict';
        var d = c(0)
          , e = c(7).sprintf
          , f = c(3)
          , g = c(1023);
        a.exports = {
            getFoo: function(k, l) {
                return _Mathround(k.score);
            },
            ...
        };
    },
    ...
});
```

with this error `TypeError: _Mathround is not a function`

Chrome devtool sees `_Mathround` as a Global variable when I try to debug, but it is `undefined`

vigneshshanmugam:
+ Deopt case was not there for classBody
+ handle the case when creating segments when the parentPath is not a function. 
+ fix #458 
+ fix #529

### DIFF
--- a/packages/babel-plugin-minify-builtins/src/index.js
+++ b/packages/babel-plugin-minify-builtins/src/index.js
@@ -148,27 +148,27 @@ function getSegmentedSubPaths(paths) {
     paths,
     (lastCommon, index, ancestries) => {
       // found the LCA
-      if (!lastCommon.isProgram() && lastCommon.isFunction()) {
-        segments.set(lastCommon, paths);
-        return;
-      }
-      // Found the LCA in the parent
-      let funParent = lastCommon.getFunctionParent();
-      if (!(lastCommon.isProgram() || funParent.isProgram())) {
-        segments.set(funParent, paths);
-        return;
+      if (!lastCommon.isProgram()) {
+        let fnParent;
+        if (lastCommon.isFunction()) {
+          segments.set(lastCommon, paths);
+          return;
+        } else if (!(fnParent = lastCommon.getFunctionParent()).isProgram()) {
+          segments.set(fnParent, paths);
+          return;
+        }
       }
 
       // Deopt and construct segments otherwise
       for (const ancestor of ancestries) {
-        const funcPath = getChildFuncion(ancestor);
-        if (funcPath === void 0) {
+        const fnPath = getChildFuncion(ancestor);
+        if (fnPath === void 0) {
           return;
         }
         const validDescendants = paths.filter(p => {
-          return p.isDescendant(funcPath);
+          return p.isDescendant(fnPath);
         });
-        segments.set(funcPath, validDescendants);
+        segments.set(fnPath, validDescendants);
       }
     }
   );
@@ -181,7 +181,6 @@ function getChildFuncion(ancestors = []) {
       return path;
     }
   }
-  return void 0;
 }
 
 function hasPureArgs(path) {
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
Eslint config is broken, but errors are not shown in the travis build.

packages/babel-plugin-minify-builtins/src/index.js
```diff
@@ -129,44 +133,62 @@ module.exports = function({ types: t }) {
...
+
+  function getChildFunction(node, wm) {
+    let funcNode;
+    if (wm.has(node)) {
+      return wm.get(node);
     }
-  );
+    t.traverseFast(node, subNode => {
+      if (t.isFunction(subNode)) {
+        funcNode = subNode;
```
vigneshshanmugam: will try this with generators if we can avoid full traversal.

vigneshshanmugam:
@boopathi This is ready for review now and all the edge cases and included in the test.

packages/babel-plugin-minify-builtins/src/index.js
```diff
@@ -147,28 +147,43 @@ function getSegmentedSubPaths(paths) {
...
   return segments;
 }
 
+function getChildFuncion(ancestors = []) {
+  for (const path of ancestors) {
+    if (path.isFunction()) {
+      return path;
+    }
+  }
+  return void 0;
```
boopathi: nit: this statement is not required

packages/babel-plugin-minify-builtins/src/index.js
```diff
@@ -147,28 +147,43 @@ function getSegmentedSubPaths(paths) {
...
+      if (!lastCommon.isProgram() && lastCommon.isFunction()) {
         segments.set(lastCommon, paths);
         return;
       }
+      // Found the LCA in the parent
+      let funParent = lastCommon.getFunctionParent();
+      if (!(lastCommon.isProgram() || funParent.isProgram())) {
+        segments.set(funParent, paths);
+        return;
+      }
```
boopathi: in both the above conditions, `lastCommon.isProgram()` should be false. So, why not combine this logic

```js
let fnParent = lastCommon.getFunctionParent();
if (
  !lastCommon.isProgram() &&
  (lastCommon.isFunction() || !lastCommon.getFunctionParent().isProgram())
) {
  segments.set(lastCommon, paths);
  return;
}
```

boopathi:
Let's also handle the case where the arrowFn body isn't a block statement. 

for example: 

```js
function foo() {
  const x = () => Math.max(Math.max(a,b), Math.max(Math.floor(b), Math.floor(c)))
}
```

vigneshshanmugam:
Good catch. Can we construct the body statement or find another child fn?

PR #539, opened 2017-05-22
Upgrade deps and update benchmarks

boopathi:


kzc:
`butternut` has a bug that skips sections of `vue.js`: https://github.com/Rich-Harris/butternut/issues/162

You may want to remove `vue.js` from the comparison until that `butternut` issue is resolved.

boopathi:
Ah! Thanks. I'll remove it.

PR #543, opened 2017-05-22
Fix removal of quotes from property names

### LINKED ISSUE #415: Syntax errors after minifying unicode

boopathi:
+ Fix #416 
+ Fix #445 
+ Fix #415

### DIFF
--- a/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
+++ b/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
@@ -88,7 +88,8 @@ describe("transform-property-literals-plugin", () => {
     const expected = unpad(`
       ({
         ಠ_ಠ: "bar",
-        "12e34": "wut"
+        "12e34": "wut",
+        \u01FC: "AE"
       });
     `);
     expect(transform(source)).toBe(expected);
--- a/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
+++ b/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js
@@ -40,7 +40,13 @@ describe("transform-property-literals-plugin", () => {
         "import": null
       });
     `);
-    expect(transform(source)).toBe(source);
+    const expected = unpad(`
+      ({
+        default: null,
+        import: null
+      });
+    `);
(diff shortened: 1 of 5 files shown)


boopathi:
@mathiasbynens Would like you to review this. There are a few more corner cases missing I suppose - for example - `.123`. But, is there a esutils API like - `isPropertyNameES5` like `isIdentifierES5` ?

boopathi:
Or would it suffice to only transform valid Identifiers (`esutils.keyword.isIdentifierNameES5`) and not care about property names ?

mathiasbynens:
There’s no `isPropertyName*` API in esutils (yet). I wonder how @Constellation and @michaelficarra feel about potentially adding it.

michaelficarra:
@mathiasbynens Are you thinking of something like this? https://github.com/shapesecurity/shift-codegen-js/blob/0d09bd8a881c62c1bb85d6ff3cfa09697867ed0a/src/minimal-codegen.js#L636-L643

mathiasbynens:
@michaelficarra Exactly!

boopathi:
@michaelficarra Thanks a lot. I used the code from that repository and have also added the license. Please review the changes.

PR #544, opened 2017-05-22
Rename package inline-consecutive-adds

boopathi:


boopathi:
THis is not required now :)

vigneshshanmugam:
:D good then

PR #553, opened 2017-05-24
Remove redundant use strict (Fix #542)

### LINKED ISSUE #542: Remove unnecessary "use strict"
If the parent has a "use strict" directive, it's not necessary to have it in the children.

boopathi:


packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Remove redundant use strict
+ * If the parent has a "use strict" directive, it is not required in
+ * the children
+ *
+ * @param {NodePath} block BlockStatement
+ */
+function removeUseStrict(block) {
+  if (!block.isBlockStatement()) {
```
vigneshshanmugam: why is this check even needed if its already checked before.

boopathi: since it is a separate file. to not shoot ourselves in foot by calling `removeUseStrict(SomeOtherPath)`

vigneshshanmugam: Cool! didn't realise it was a separate module.

Issue #538, opened 2017-05-21
Cleanup when smoke tests fail


boopathi:
1. Run smoke tests (when we have a failing smoke test)
2. The minified bundle in the asset is wrong
3. Re-run smoke test - now this tries to minify the already wrongly minified file.

So, we need to clean-up after it has failed before exit.

PR #557, opened 2017-05-25
Clean up before every run to avoid working on already minified file

buunguyen:
Fixes https://github.com/babel/babili/issues/538.

Approach taken: perform a cleanup before a every run.  Alternative: cleanup at failure. 

I chose the former because:

* It's more robust. There could be interruption other than failure.
* Upon failure, (I guess) the minified file might be useful for troubleshooting.


boopathi:
Thanks.

PR #567, opened 2017-06-06
Add workaround for Safari for loop lexical scope bug

### LINKED ISSUE #559: Cannot redeclare let variable in Safari
After upgrading to babili 0.1.2 following error appears in Safari 10.1 (running ES6 native): `SyntaxError: Cannot declare a let variable twice: 'e'.`. It happens on fragments like this (after minify I have a lot of this): 

```js
function a(e) {
  let e = 1;
  console.log(e);
}
a(2);

// Real life minify result:

  function Ge(e, t) {
    const n = e.toLowerCase(),
          a = n.split(/\s+/),
          r = [];

    for (let e = 0, n = 0; n < t && e < a.length; n += a[e].length, e++) r.push(a[e].split('').map((e) => e.normalize('NFKD').charAt(0)).filter((e) => /[A-z]/.test(e)).join(''));
    return r.join('-');
  }
```

and it seems for me that Safari counts function parameters as declared by `let`. Neither Chrome, Firefox or Edge have such problem. Is there any workaround apart of switching off minify (as I'm forced to do now)?

btmills:
Safari raises a syntax error for a `let` or `const` declaration in a `for` loop initalization that shadows a parent function's parameter:

```js
function a(b) {
  for (let b;;);
}
```

This is valid code, but Safari throws a syntax error. The bug has been [reported](https://bugs.webkit.org/show_bug.cgi?id=171041) and since [fixed](https://trac.webkit.org/changeset/217200/webkit/trunk/Source) in WebKit, so future versions of Safari will not have this problem.

This modifies the scope tracker's `canUseInReferencedScopes` method to detect cases where a rename would trigger this bug and use a different name.

Fixes #559 

This is my first time dealing with Babel internals, so please let me know if there's a better way to do this!

### DIFF
--- a/packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
+++ b/packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
@@ -126,27 +126,31 @@ module.exports = class ScopeTracker {
     // https://github.com/babel/babili/issues/559
     // https://bugs.webkit.org/show_bug.cgi?id=171041
     // https://trac.webkit.org/changeset/217200/webkit/trunk/Source
-    const isBlockScoped = binding.kind === "let" || binding.kind === "const";
-    const isForLoopDeclaration =
-      (binding.path.parentPath.parent.type === "ForStatement" &&
-        binding.path.parentPath.key === "init") ||
-      (binding.path.parentPath.parent.type === "ForInStatement" &&
-        binding.path.parentPath.key === "left") ||
-      (binding.path.parentPath.parent.type === "ForOfStatement" &&
-        binding.path.parentPath.key === "left");
-    if (isBlockScoped && isForLoopDeclaration) {
-      const functionParentScope = binding.scope.getFunctionParent();
-      const isIncorrectlyTopLevelInSafari =
-        binding.scope.parent === functionParentScope;
-      if (isIncorrectlyTopLevelInSafari) {
-        const parentFunctionBinding = this.bindings
-          .get(functionParentScope)
-          .get(next);
-        if (parentFunctionBinding) {
-          const parentFunctionHasParamBinding =
-            parentFunctionBinding.kind === "param";
-          if (parentFunctionHasParamBinding) {
-            return false;
+    const maybeDecl = binding.path.parentPath;
+    const isBlockScoped =
+      maybeDecl.isVariableDeclaration({ kind: "let" }) ||
+      maybeDecl.isVariableDeclaration({ kind: "const" });
+    if (isBlockScoped) {
+      const maybeFor = maybeDecl.parentPath;
+      const isForLoopBinding =
+        maybeFor.isForStatement({ init: maybeDecl.node }) ||
+        maybeFor.isForXStatement({ left: maybeDecl.node });
+      if (isForLoopBinding) {
+        const fnParent = maybeFor.getFunctionParent();
+        if (fnParent.isFunction()) {
+          const parentScope = maybeFor.scope.parent;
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
```diff
@@ -121,6 +121,37 @@ module.exports = class ScopeTracker {
...
+    const isBlockScoped = binding.kind === "let" || binding.kind === "const";
+    const isForLoopDeclaration =
+      (binding.path.parentPath.parent.type === "ForStatement" &&
+        binding.path.parentPath.key === "init") ||
+      (binding.path.parentPath.parent.type === "ForInStatement" &&
+        binding.path.parentPath.key === "left") ||
+      (binding.path.parentPath.parent.type === "ForOfStatement" &&
+        binding.path.parentPath.key === "left");
+    if (isBlockScoped && isForLoopDeclaration) {
+      const functionParentScope = binding.scope.getFunctionParent();
```
boopathi: thing to note here: `getFunctionParent` in babel returns either a `Function` or `Program`. Is this okay? or de we specifically look for function?

eg:

```js
let c;
```

packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
```diff
@@ -121,6 +121,37 @@ module.exports = class ScopeTracker {
...
+    // https://bugs.webkit.org/show_bug.cgi?id=171041
+    // https://trac.webkit.org/changeset/217200/webkit/trunk/Source
+    const isBlockScoped = binding.kind === "let" || binding.kind === "const";
+    const isForLoopDeclaration =
+      (binding.path.parentPath.parent.type === "ForStatement" &&
+        binding.path.parentPath.key === "init") ||
+      (binding.path.parentPath.parent.type === "ForInStatement" &&
+        binding.path.parentPath.key === "left") ||
+      (binding.path.parentPath.parent.type === "ForOfStatement" &&
+        binding.path.parentPath.key === "left");
```
boopathi: ForIn and ForOf can be combined as `ForXStatement` and it's better to use path's API to verify by passing the node - 

```js
path.isForStatement({ init: varDeclaration.node });
path.isForXStatement({ left: varDeclaration.node });
```

Wrote something up with a few more corner cases here - http://astexplorer.net/#/gist/1336be60e031cffc194b693ec358074f/c8d813db7e9ff40eb57be5ac807ce16751554fd8

boopathi:
Thanks for the PR. It looks great. I have added some review inline. I'll try this out tomorrow.

packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
btmills: Ah, good to know. Safari appears fine with `let a; for (let a; false;);` - no syntax error there. I found `.isFunction()` and used that to make sure that we're inside a function scope.

packages/babel-plugin-minify-mangle-names/src/scope-tracker.js
btmills: Thanks for the tips! I've updated this PR with what I hope is more idiomatic traversal code. The only intentional change to the logic is ensuring that the parent scope belongs to a function and not `Program`.

I also added a whole bunch of other tests including the other corner cases you identified.

jason-krypton:
Any eta on this PR going in and being released? Old babili releases have some breaking bugs and this is blocking us from upgrading.

alicoding:
I just encountered this one as well :)

Issue #422, opened 2017-02-12
Simplify after DCE (Multiple Passes)


boopathi:
Related #61 

Right now DCE is run in a single pass at the Program.exit. But this blocks other optimisations that could be performed after DCE - such as simplify. 

an example - from #420  ... if_return optimisation (simplify) after DCE is not happening here because DCE is at the end, and requires a second pass. 

Also, if the function is an arrow function, after DCE, an arrow function with a single return statement could be further simplified to remove the BlockStatement, which is also not happening now because of the same reasons.

```js
// input
function foo(...args) {
  if (!i) return;
  else {
    const x = {args};
    while(false);
    return x;
  }
};

// out
function foo(...a) {
  if (!i) return;
  return { args: a };
}

// could reduce to
function foo(...a) {
  if (i) return {args: a};
}
```

Issue #61, opened 2016-07-29
Multiple passes?


boopathi:
Are there situations that we might need multiple passes for better minification ?

For example, one plugin might transform

``` js
if (a) foo (x) : foo (y);
//to
a ? foo (x) : foo (y)
```

and another plugin might transform 

``` js
//to
foo (a ? x : y);

// or even,
x ? foo(a) : foo(b)
//to
foo (x ? a : b)
```

and another one could remove this as pure expression statement. or modify it differently. I'm not able to think about it.

So the order of plugins in the preset will matter, and we say (with plugins- a and b) the order is `a, b`, and another particular transformation might require the order to be `b, a`. So, should we explore multiple passes. I'm not sure if this is a problem. Just asking if there are situations that you faced like this.

boopathi:
#317 Simplify and DCE order

kangax:
A brute force solution would be to add a preset option that simply does 2 passes. When I tested this a while ago I also saw that sometimes 3 passes can increase size even more. Not sure how to deal with this. How to strike balance between transformation time and size, where a tiny improvement can cause 2x or 3x longer times.

boopathi:
> 3 passes can increase size even more

this should be the mangler

boopathi:
#422 Simplify after DCE

vigneshshanmugam:
lets take advantage of the plugin ordering option https://github.com/babel/babel/pull/5735

PR #572, opened 2017-06-08
Replace `x != undefined` with `x != null`

j-f1:
~Bonus: replace `undefined == null` and `null != null` with `!0` and `!1`, respectively.~

Tests included.


packages/babel-plugin-transform-simplify-comparison-operators/__tests__/eqnull-test.js
```diff
@@ -0,0 +1,44 @@
+jest.autoMockOff();
+
+const babel = require("babel-core");
+const plugin = require("../src/index");
+const unpad = require("../../../utils/unpad");
```
vigneshshanmugam: unused.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -9,6 +14,17 @@ module.exports = function() {
       BinaryExpression(path) {
         const { node } = path;
         const op = node.operator;
+        const negated = node.operator.startsWith("!");
+ 
+        if (["!=", "=="].includes(node.operator)) {
```
vigneshshanmugam: use indexOf, we still support node 4

packages/babel-plugin-transform-simplify-comparison-operators/__tests__/eqnull-test.js
j-f1: Done.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
j-f1: Fixed.

packages/babel-plugin-transform-simplify-comparison-operators/__tests__/eqnull-test.js
```diff
@@ -0,0 +1,43 @@
...
+    const expected = "x == null;";
+    expect(transform(source)).toBe(expected);
+  });
+  it("should not shorten `x === undefined`", () => {
+    const source = "x === undefined;";
+    expect(transform(source)).toBe(source);
+  });
+
+  it("should shorten `undefined == null`", () => {
+    const source = "undefined == null";
```
boopathi: this should be already taken care of by constant-folding plugin. This plugin should not worry about these folding constants.

j-f1: It [doesn’t appear to handle this](http://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=react%2Cstage-0&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&code=f(undefined%20%3D%3D%20null)&playground=true). Are you saying that I should move it there?

boopathi: That can be fixed in `path.evaluate` of babel if that doesn't handle it. Or it's an ordering issue of who runs first - #61 #422.

j-f1: I could add `void 0` to `isNull`.

j-f1: How does that sound @boopathi?

boopathi: I meant this entire replacement shouldn't be a part of this plugin. It should be handled in `path.evaluate` of babel.

j-f1: @boopathi Done. Do you know where the source for `path.evaluate` is so I can open a PR there?

xtuc:
I'm wondering if we can do this, isn't it unsafe? `null` and `undefined` are not equal.

j-f1:
@xtuc `null` and `undefined ` are double-equals equal, not triple-equals equal.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -9,6 +9,13 @@ module.exports = function() {
       BinaryExpression(path) {
         const { node } = path;
         const op = node.operator;
+
+        if (["!=", "=="].indexOf(node.operator) !== -1) {
+          if (t.isIdentifier(node.right, { name: "undefined" })) {
```
boopathi: `void 0` will not be transformed.

boopathi:
Can you add some tests for this ?

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
j-f1: Fixed.

j-f1:
@boopathi Are you assuming in babili that people won’t do `void someFunctionWithSideEffects()`?

j-f1:
@boopathi Tests restored.

boopathi:
> @boopathi Are you assuming in babili that people won’t do `void someFunctionWithSideEffects()`?

Constant-folding plugin should take care of folding constants. `void foo()` should be a deopt anyway here. It should check only for `void x` where `pathOfX.isPure()` is `true`.

j-f1:
@boopathi This is ready for review.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -9,6 +15,16 @@ module.exports = function() {
       BinaryExpression(path) {
         const { node } = path;
         const op = node.operator;
+
+        if (["!=", "=="].indexOf(node.operator) !== -1) {
+          if (
+            t.isIdentifier(node.right, { name: "undefined" }) ||
+            isPureVoid(path.get("right"))
```
boopathi: This will work only for `void right`. What about `undefined == foo` ?

j-f1: @boopathi Fixed.

boopathi:
Just a few thoughts on this -

1. Should it go into comparison-operators?
1. Since this is an unsafe transformation, it's better to put this entire transformation behind an option - `undefinedToNull` or `voidToNull` or anything else relevant. 
1. Update tests to use `thePlugin`
1. `t.isIdentifier("undefined")` check should also check for bindings in the scope `path.scope.getBinding("undefined")` there is a defined `undefined`.

j-f1:
@boopathi 

1. 👍
2. I don’t think it’s unsafe. Why do you think it’s unsafe?
3. 👍
4. 👍

xtuc:
2) it's safe since https://github.com/babel/minify/pull/572#issuecomment-307738921 (see [Equality_comparisons_and_sameness](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness))

j-f1:
@boopathi 

1. It’s already in comparison-operators.
2. ❓
3. ✅
4. ✅

boopathi:
My question is whether it should go into comparison operators, or should it go into simplify. IMO, Comparison operators simplifies the operator used without changing the operands.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -1,6 +1,25 @@
...
+    return (
+      path.isIdentifier({ name: "undefined" }) &&
+      !path.scope.getBinding("undefined")
+    );
+  }
+
+  function undefinedToNull(path) {
+    if (isRealUndefined(path) || isPureVoid(path)) {
+      path.replaceWith(t.nullLiteral());
+      return true;
```
vigneshshanmugam: is this return true / false needed? I don't see it used anywhere for conditional check.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -1,6 +1,25 @@
...
-module.exports = function() {
+module.exports = function({ types: t }) {
+  function isPureVoid(path) {
+    return path.isUnaryExpression({ operator: "void" }) && path.isPure();
+  }
+
+  function isRealUndefined(path) {
+    return (
+      path.isIdentifier({ name: "undefined" }) &&
+      !path.scope.getBinding("undefined")
```
vigneshshanmugam: why is this check required?. Even if someone does 
```js
var undefined = 'blah';
```

Its a read only property and does not impact our null conversion.

j-f1: @boopathi said:

> `t.isIdentifier("undefined")` check should also check for bindings in the scope `path.scope.getBinding("undefined")` there is a defined `undefined`.

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
j-f1: Removed 🔥

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
vigneshshanmugam: /cc @boopathi Any reason why the check is needed? Am i missing some edge case?

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
vigneshshanmugam: 👍

packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
boopathi:
```js
function foo(undefined) { return undefined == void 0 } 
foo("bar"); // true or false ?
```

vigneshshanmugam: `true` even if undefined or both are converted to null

j-f1:
```js
> function foo(undefined) { return undefined == void 0 } 
> foo("bar"); // true or false ?
// false
```

vigneshshanmugam: okay got it 👍

boopathi:
It's not immediately obvious that this goes into `comparison-operators` plugin. This question still applies - 

> My question is whether it should go into comparison operators, or should it go into simplify. IMO, Comparison operators simplifies the operator used without changing the operands

j-f1:
@boopathi If you’d like me to move it to `simplify`, I’ll gladly do so. Should I?

j-f1:
Sorry for the delay on this @boopathi 😢

I’ve moved the changes to `babel-plugin-minify-simplify`.

j-f1:
Looks like the Travis failure is unrelated. Closing and reopening to restart the build.

j-f1:
Ping @boopathi and @vigneshshanmugam for 👀

boopathi:
@j-f1 Hi, Thanks for the PR. Right now let's concentrate on fixing the existing bugs, babel-7, improving the speed of transformations and fixing the memory issue to get it to 1.0. New set of features may delay that. I'll leave it open and postpone it to sometime later.

PR #578, opened 2017-06-14
Remove folding Array.prototype.concat (fix #577)

### LINKED ISSUE #577: Incorrect folding of Array.concat (babel-plugin-minify-constant-folding)
I've noticed a strange behavior due to the minify-constant-folding transform used in babili >= 0.1.0.

Test case:

```js
var a = ['first'];
var b = ['second'].concat(a);
// babel-preset-babili >= 0.1.0 produces:
// var b=['second',a];			(INCORRECT)

// babel-preset-babili <= 0.0.12 leaves untouched:
// var b=['second'].concat(a);		(CORRECT)

if (b[0] !== 'second' || b[1] !== 'first') {
	throw new Error('fail!')
} else {
	console.log('pass!');
}
```

Here's an alternate case, based on the example found in the package's [README](https://github.com/babel/babili/blob/master/packages/babel-plugin-minify-constant-folding/README.md):
```js
// README Example:
[a, b, c].concat([d, e], f, g, [h]);
// becomes [a, b, c, d, e, f, g, h]; 		(CORRECT)

// Functionally equivalent:
var x = [d, e];
[a, b, c].concat(x, f, g, [h]);
// becomes [a, b, c, x, f, g, h]; 		(INCORRECT)
```

For reference, here's my .babelrc I used to test/produce the above results:
```js
{
    "presets": [],
    "minified": true,
    "compact": true,
    "sourceMaps": true,
    "retainLines": true,
    "plugins": [
        "minify-constant-folding",
    ],
}
```
Let me know if you'd like any further information.

Thanks,
Aphix

boopathi:


packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
```diff
@@ -107,9 +107,6 @@ describe("constant-folding-plugin", () => {
   it("should handle Array methods on array literals", () => {
     const source = unpad(
       `
-      [1, 2, 3].concat([4, 5, 6]);
-      [a, b, c].concat([d, e], f, g, [h]);
-      [1, 2, 3]["concat"]([4, 5, 6]);
```
j-f1: You might want to test this with `push` since you’re removing it:
```js
[1, 2, 3]["push"]([4, 5, 6])
```

packages/babel-plugin-minify-constant-folding/src/replacements.js
```diff
@@ -32,16 +32,6 @@ module.exports = ({ types: t }) => {
         }
       },
       calls: {
-        concat(...args) {
-          return t.arrayExpression(
-            this.elements.concat(
-              ...args.map(arg => {
-                if (t.isArrayExpression(arg)) return arg.elements;
```
j-f1: Would it be possible to check if any of these are variables instead of removing this entirely? For example, `[a, b, c].concat([1, 2, 3])` is safe.

boopathi: Is it really common to use constants here in concat ?

packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
boopathi: Done 👍

packages/babel-plugin-minify-constant-folding/src/replacements.js
j-f1: I guess not.

PR #587, opened 2017-06-16
Fix #576

### LINKED ISSUE #576: Don't use path.replaceWith() with a source string, use path.replaceWithSourceString()
ERROR in unknown: Don't use `path.replaceWith()` with a source string, use `path.replaceWithSourceString()`

Seems to be the result of this change : 
https://github.com/babel/babili/pull/543/files

j-f1:
People were calling `"foo".toString()` or `["foo", "bar"].toString()`, which returned the real JS string `"[object Object]"`, the stringified version of the replacements object. Fixes #576.

---

I just noticed this:
```js
[foo, bar].toString() // is longer than
[foo, bar].join(',')
```

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/constant-folding-test.js
@@ -237,4 +237,18 @@ describe("constant-folding-plugin", () => {
     );
     expect(transform(source)).toBe(expected);
   });
+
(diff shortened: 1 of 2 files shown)


boopathi:
Thanks. Can you also add a test for this ?

j-f1:
@boopathi Test added.

damianobarbati:
any chance to merge this in master soon? production build broken for many! 😱

vigneshshanmugam:
We will release a new version today with this fix.

Issue #48, opened 2016-07-19
if..else dead_code elimination - preserve vars


boopathi:
``` js
if ( 0 ) {
  var x = foo();
}

// or
if (0) var x = bar();
```

out:

``` js
var x;
```

This can either be done by hoisting vars for all functions before IfStatement is reached, or by analysing the pathToBeRemoved for var declarations.

I tried the latter approach here - https://github.com/boopathi/babel-minify/blob/0fbb45d18b13df53dd225dbd36ef0af88c7190ea/packages/babel-plugin-transform-conditionals/src/replace-with.js#L65-L88

PR #597, opened 2017-06-21
[fix #574] deopt when binding is present in diff scope

### LINKED ISSUE #574: if statement wrongly removed when var declaration inside conditional block
It seems that when you declare and initialize a variable in a code block that executes conditionally, it is assumed that the variable will have this value also outside of the code block. This is wrong when the conditional code block does not get executed, in which case the variable will be undefined. As a result, further optimizations, like removing if statements, produce wrong results.

E.g., the following code produces a wrong result

```js
function x(v) {
  if (v) var w = true;
  if (w) console.log("hello",v);
}

x(false);
x(true);
```

It produces this:

```js
function x(a){if(a);console.log("hello",a)}x(!1),x(!0);
```

Which, when executed, outputs:

```
hello false
hello true
```

while it should only output the second line.

A related problem is with this code:

```js
function x(v) {
  if (v) {var w = [];for (let i=0;i<1;i++) w.push("hello");}
  if (w) for (let i=0;i<w.length;i++) console.log(w[i],v);
}

x(false);
x(true);
```

Which produces:

```js
function x(a){if(a){var b=[];for(let a=0;1>a;a++)b.push("hello")}for(let c=0;c<b.length;c++)console.log(b[c],a)}x(!1),x(!0);
```

When executed this throws an exception `Cannot read property 'length' of undefined`.

However, unlike the previous example, when disabling the `minify-dead-code-elimination` plugin, this produces correct code.

The latter example is what happens for example when processing the firebase scripts (see: https://github.com/Polymer/polymer-cli/issues/701)

vigneshshanmugam:
+ Fix #574 

this is required for only var bindings. 

The other way we could fix this is to hoist all the var bindings before we run simplify & deadcode.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -2462,4 +2462,14 @@ describe("dce-plugin", () => {
       }
     `
   );
+
+  thePlugin(
+    "should deopt when binding is on different scope - issue #574",
+    `
+      function foo(v) {
+        if (v) var w = 10;
+        if (w) console.log("hello", v);
+      }
+    `
+  );
 });
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -776,16 +776,15 @@ module.exports = ({ types: t, traverse }) => {
           const evalResult = test.evaluate();
           const isPure = test.isPure();
 
-          const { path: bindingPath } =
-            path.scope.getBinding(test.node.name) || {};
+          const binding = path.scope.getBinding(test.node.name);
 
           // Ref - https://github.com/babel/babili/issues/574
           // deopt if var is declared in other scope
           // if (a) { var b = blahl;} if (b) { //something }
           if (
-            bindingPath &&
-            bindingPath.parentPath.isVariableDeclaration() &&
-            isDifferentScope(bindingPath.parentPath, path)
+            binding &&
+            binding.path.parentPath.isVariableDeclaration() &&
+            isDifferentScope(binding.path.parentPath, path)
           ) {
             return;
           }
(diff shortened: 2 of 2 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -285,14 +291,10 @@ module.exports = ({ types: t, traverse }) => {
               let replacementPath = binding.path;
               let isReferencedBefore = false;
 
-              if (binding.referencePaths.length > 1) {
```
vigneshshanmugam: unnecessary check.

packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -765,8 +767,21 @@ module.exports = ({ types: t, traverse }) => {
           const evalResult = test.evaluate();
           const isPure = test.isPure();
 
-          const replacements = [];
+          const { path: bindingPath } =
+            path.scope.getBinding(test.node.name) || {};
```
boopathi: don't use `|| {}` ?

boopathi: `if (binding)` check is clearer, like it's used everywhere else in babili

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2710,4 +2710,14 @@ describe("dce-plugin", () => {
     `);
     expect(transform(source)).toBe(expected);
   });
+
+  it("should deopt when binding is on different scope - issue #574", () => {
+    const source = unpad(`
+      function foo(v) {
+        if (v) var w = 10;
+        if (w) console.log("hello", v);
```
boopathi: There is also

```js
while (v) { var w = 10; }
if (w) console.log(w);
```

vigneshshanmugam:
We need to handle this in simplify as well for the same case. 

Started to think, hoisting that var would be a better option than monkey patching at all places. Thoughts? @boopathi

boopathi:
var hoisting was considered earlier. But I didn't try measuring its effect on gzip yet. https://github.com/babel/babili/issues/48#issuecomment-233697001

vigneshshanmugam:
Hmm alright, the problem itself is with babel's `path.evaluate()` being truthy even for bindings present in different scope. 

It becomes tricky to do the scope check on every pattern matching step in simplify.

PR #602, opened 2017-06-23
fix(minify-dead-code): fix #601

### LINKED ISSUE #601: invalid block directive in minify-dead-code-elimination
Just encountered this error (w/ `babel-plugin-minify-dead-code-elimination@0.1.7`):

```
TypeError: bundle.js: block.get(...).filter is not a function
    at getUseStrictDirectives (.../node_modules/babel-plugin-minify-dead-code-elimination/lib/remove-use-strict.js:50:34)
    at removeUseStrict (.../node_modules/babel-plugin-minify-dead-code-elimination/lib/remove-use-strict.js:21:20)
    at PluginPass.exit (.../node_modules/babel-plugin-minify-dead-code-elimination/lib/index.js:844:13)
    at newFn (.../node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (.../node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (.../node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (.../node_modules/babel-traverse/lib/path/context.js:117:8)
    at TraversalContext.visitQueue (.../node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitSingle (.../node_modules/babel-traverse/lib/context.js:108:19)
    at TraversalContext.visit (.../node_modules/babel-traverse/lib/context.js:192:19)
    at Function.traverse.node (.../node_modules/babel-traverse/lib/index.js:114:17)
```

I've added some debug traces and it seems that `block.get("directives")` doesn't always return an array, but sometimes an object too, in which case this error is triggered. Not sure about the internals of blocks, but maybe that's an upstream issue?

As a quick fix this works:

```js
function getUseStrictDirectives(block) {
    var dir = block.get("directives");
    return dir.filter ? dir.filter(function (directive) {
        return directive.node.value.value === useStrict;
    }) : [];
}
```

postspectacular:
- add block checks to getUseStrictDirectives()

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
@@ -50,7 +50,8 @@ function isStrict(block) {
 }
 
 function getUseStrictDirectives(block) {
-  return block.get("directives").filter(directive => {
-    return directive.node.value.value === useStrict;
-  });
+  return block.body && block.body.length > 0 ?
+    block.get("directives").filter(function (directive) {
+      return directive.node.value.value === useStrict;
+  }) : [];
 }
--- a/packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
@@ -50,8 +50,8 @@ function isStrict(block) {
 }
 
 function getUseStrictDirectives(block) {
-  return block.body && block.body.length > 0 ?
-    block.get("directives").filter(function (directive) {
-      return directive.node.value.value === useStrict;
+  var dir = block.get("directives");
+  return dir.filter ? dir.filter(function (directive) {
+    return directive.node.value.value === useStrict;
   }) : [];
 }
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/remove-use-strict.js
```diff
@@ -50,7 +50,10 @@ function isStrict(block) {
 }
 
 function getUseStrictDirectives(block) {
-  return block.get("directives").filter(directive => {
-    return directive.node.value.value === useStrict;
-  });
+  var dir = block.get("directives");
```
vigneshshanmugam: why not check for array? `Array.isArray()`

postspectacular: fixed

boopathi:
Can you also add a test case for this ?

postspectacular:
@boopathi there already is a test for this. I still don't know how to produce a minimal example of the case triggering the original error. The existing test does succeed for the fix, though and the fix also allows me to process my project codebase now...

PR #606, opened 2017-06-29
Refactor the test suite

j-f1:
This adds a test helper that creates a testing function, which is usually called `thePlugin`, which streamlines testing transforms. I’ve also gone through and edited some of the test names for clarity.

### DIFF
--- a/utils/test-transform.js
+++ b/utils/test-transform.js
@@ -1,0 +1,78 @@
+const babel = require("babel-core");
+
+const unpad = require("./unpad");
+
+function _transform(source, options) {
+  return babel.transform(unpad(source), options).code.trim();
+}
+
+function makeTester(
+  plugins,
+  opts,
+  { transform = _transform, check },
+  excludeKeys = []
+) {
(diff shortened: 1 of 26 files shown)


utils/test-transform.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+      } catch (e) {
+        // use the stack from outside the it() clause
+        // (the one inside the clause doesn’t show the actual test code)
+        e.stack = stack;
+        throw e;
+      }
+    });
+  };
+  thePlugin.skip = name => it.skip(name);
```
vigneshshanmugam: you would also need to support `only` if we would like to run a single test during dev..

vigneshshanmugam:
The cleanup looks good overall.. Thanks a lot 👍

utils/test-transform.js
j-f1: Done!

boopathi:
This looks great!

I thought of moving it to fixtures based testing - each test case in 2 files - `source.js` and `expected.js`. But this is fine for now.

PR #613, opened 2017-07-02
constant-folding: Fix some literal types used in `.join()`, closes #612

### LINKED ISSUE #612: constant-folding template strings
There is a bug in the [babel-plugin-minify-constant-folding](https://github.com/babel/babili/tree/master/packages/babel-plugin-minify-constant-folding) package where it incorrectly inlines template strings. Here is a simple example:

```
const a = ["a", "b"].join(' ');
// -> const a="a b";


const b = [`a${a}`, `b${a}`].join(' ');
// -> const b=" ";
```

In the second example it should keep the interpolation and produce a template string, but it does not, instead returning just an empty string. The result is broken code.

goto-bus-stop:
The `.join()` inlining function was checking for `isLiteral()` on each array element, and then using the element's `.value`. Some literals don't have `.value`s though: null literals, regexp literals and template literals.

This patch fixes `.join()`ing regexp literals and bails out of joining template literals entirely. 

Null literals already did the right thing despite not having a `.value`, because `.join()` appears to treat `undefined` and `null` the same.

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/src/replacements.js
+++ b/packages/babel-plugin-minify-constant-folding/src/replacements.js
@@ -56,6 +56,9 @@ module.exports = ({ types: t }) => {
               ) {
                 return el.value;
               }
+              if (t.isTemplateLiteral(el) && el.expressions.length === 0) {
+                return el.quasis[0].value.cooked;
+              }
               bad = true;
               return;
             })
--- a/packages/babel-plugin-minify-constant-folding/src/replacements.js
+++ b/packages/babel-plugin-minify-constant-folding/src/replacements.js
@@ -43,11 +43,21 @@ module.exports = ({ types: t }) => {
           let bad = false;
           const str = this.elements
             .map(el => {
-              if (!t.isLiteral(el)) {
-                bad = true;
-                return;
+              if (t.isRegExpLiteral(el)) {
+                return `/${el.pattern}/${el.flags}`;
               }
-              return el.value;
+              if (t.isNullLiteral(el)) {
+                return null;
+              }
+              if (
+                t.isStringLiteral(el) ||
+                t.isBooleanLiteral(el) ||
+                t.isNumericLiteral(el)
+              ) {
+                return el.value;
+              }
+              bad = true;
+              return;
             })
             .join(sep.value);
           return bad ? undefined : t.stringLiteral(str);
--- a/packages/babel-plugin-minify-constant-folding/src/replacements.js
+++ b/packages/babel-plugin-minify-constant-folding/src/replacements.js
@@ -14,7 +14,7 @@ module.exports = ({ types: t }) => {
   function defaultZero(cb) {
     return function(i = t.numericLiteral(0), ...args) {
       if (t.isNumericLiteral(i)) {
-        return cb.call(this, this, i.value, ...args);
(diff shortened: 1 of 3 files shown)


goto-bus-stop:
Lint is failing, but if I `npm run fix` some files that are not relevant to this PR get changed too. Should I commit all of that or just the fixes to the files I changed?

j-f1:
@goto-bus-stop What about people using template strings like `` `foo` ``? You could check if `templateLiteral.expressions.length === 0`, which would indicate that it’s an ordinary template literal.

goto-bus-stop:
Ah yes, that's quite simple. Thanks for the suggestion!

packages/babel-plugin-minify-constant-folding/src/replacements.js
```diff
@@ -43,11 +43,24 @@ module.exports = ({ types: t }) => {
...
           const str = this.elements
             .map(el => {
-              if (!t.isLiteral(el)) {
-                bad = true;
-                return;
+              if (t.isRegExpLiteral(el)) {
+                return `/${el.pattern}/${el.flags}`;
               }
-              return el.value;
+              if (t.isNullLiteral(el)) {
```
vigneshshanmugam: IMO, we can do `path.evaluate(el)` and deopt. 

If its gonna be kill performance which in most cases should not be a problem since the path would already be cached during the traversal phase.. Thoughts?

packages/babel-plugin-minify-constant-folding/src/replacements.js
```diff
@@ -43,11 +43,24 @@ module.exports = ({ types: t }) => {
...
+              }
+              if (
+                t.isStringLiteral(el) ||
+                t.isBooleanLiteral(el) ||
+                t.isNumericLiteral(el)
+              ) {
+                return el.value;
+              }
+              if (t.isTemplateLiteral(el) && el.expressions.length === 0) {
+                return el.quasis[0].value.cooked;
```
vigneshshanmugam: check this https://github.com/babel/babel/blob/7.0/packages/babel-traverse/src/path/evaluation.js#L391

packages/babel-plugin-minify-constant-folding/src/replacements.js
goto-bus-stop: You mean doing `evaluate` instead of all the `is*` checks?

mgol:
This has just hit me in production. A very simple test case:
```js
const a = 'a';
console.log([`${ a }b`, 'c',].join(''));
```
transpiles to:
```js
const a='a';console.log('c');
```
instead of the expected:
```js
const a='a';console.log(`${ a }bc`);
```
or:
```js
console.log('abc');
```

boopathi:
Thanks! Looks good to me. Can you address the pending comments?

goto-bus-stop:
Added a commit to use `path.evaluate()`. Previously the NodePath was not available inside replacement functions, only the node itself. I changed it to pass the NodePath to replacement functions, and changed all existing replacement functions to use `this.node` instead of just `this` where necessary. It makes the diff quite noisy.

boopathi:
LGTM. Looks like formatting issues on CI. you can fix it with `yarn format`

goto-bus-stop:
Oops, should've checked lint before committing. Thanks for the heads-up

PR #616, opened 2017-07-03
Convert simple arrow function body to expression

ashsearle:
Eliminate unnecessary `return` keyword by converting arrow function block statements to expressions where possible.

Example:
```js
const foo = () => {
  return (1, 2);
};
const bar = () => {
  return [];
};

// becomes
const foo = () => (1, 2);
const bar = () => [];
```

### DIFF
--- a/packages/babel-plugin-minify-simplify/src/index.js
+++ b/packages/babel-plugin-minify-simplify/src/index.js
@@ -782,7 +782,6 @@ module.exports = ({ types: t }) => {
 
           if (
             t.isArrowFunctionExpression(parent) &&
-            node.body &&
             node.body.length === 1 &&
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-simplify/src/index.js
```diff
@@ -780,6 +780,17 @@ module.exports = ({ types: t }) => {
         exit(path) {
           const { node, parent } = path;
 
+          if (
+            t.isArrowFunctionExpression(parent) &&
+            node.body &&
```
boopathi: We are in a `BlockStatement` - is this check every `falsey` ? A BlockStatement's body is always an array.

Issue #585, opened 2017-06-15
Fix removal of quotes from property names


k-dauda:
Quotes on property names are still being stripped in IE in production with v0.1.3 despite this fix: https://github.com/babel/babili/pull/543

boopathi:
Right now the quotes if they can be removed from the property names according to ES5, then it will be removed.

What version of IE are you targeting?

And do you have a sample code snippet (or) the property name which is invalid in IE without quotes?

k-dauda:
@boopathi IE11 and Edge in production have this problem, an example can be seen in this issue: https://github.com/cibernox/ember-power-select/issues/915

anhpha:
will be fixed?

j-f1:
@k-dauda Are you using the [`transform-es3-property-literals`](http://babeljs.io/docs/plugins/transform-es3-property-literals/) plugin? This *should* undo the conversion if you run it after Babili.

k-dauda:
@j-f1, no I am not using that plugin

boopathi:
Fixed in #730

PR #620, opened 2017-07-07
Disable property litterals package by default

alexBaizeau:
I'm suggesting to disable the `babel-plugin-transform-property-literals` by default as it is not working correctly in ES5.

I was intending to write a failing unit test, but it turns out that there is already a test that is demonstrating a bad behaviour

https://github.com/babel/babili/blob/master/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js#L39


boopathi:
related #585

boopathi:
Going to close this one. We will fix it rather than disabling the entire plugin. There is an option for you to disable the plugin.

PR #622, opened 2017-07-08
[DCE] Fix removal of ArrayPattern

### LINKED ISSUE #615: Another "Cannot read property 'end' of null" error
### Input ###

```js
function foo(arg) {
    const { y } = arg, { z, x = null } = y;
    return z ? z : null;
}
```

### Output ###
```
TypeError: test3.js: Cannot read property 'end' of null
    at _evaluate (C:\Users\John\Documents\GitHub\babili\packages\babili\node_modules\babel-traverse\lib\path\evaluation.js:159:57)
    at evaluate (C:\Users\John\Documents\GitHub\babili\packages\babili\node_modules\babel-traverse\lib\path\evaluation.js:65:17)
    ...
```
[REPL](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2016&targets=&browsers=&builtIns=false&debug=false&code_lz=GYVwdgxgLglg9mABMOcAUBDATgcwJSIDeAsAFCIWIQIDOURiAnogL6IC8i2OANAwF58AHh0RgQAGwmtRjANxlKiLAFMoILEn6IA_Im0AuMZIkLSLIA)

The following version does not produce an exception but generates incorrect output:
### Input ###

```js
function foo(arg) {
    const { y } = arg, { z, x = null } = y;
    return z;
}
```

### Output ###
```js
function foo(a){const{y:b}=a;return z}
```
[REPL](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2016&targets=&browsers=&builtIns=false&debug=false&code_lz=GYVwdgxgLglg9mABMOcAUBDATgcwJSIDeAsAFCIWIQIDOURiAnogL6IC8i2OANAwF58AHh0RgQAGwmtRjANxlKiLAFMoILEn4LSLIA)

Note: removing `, x = null` produces correct output. Alternatively returning `x + z` produces correct output ([REPL](https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2016&targets=&browsers=&builtIns=false&debug=false&code_lz=GYVwdgxgLglg9mABMOcAUBDATgcwJSIDeAsAFCIWIQIDOURiAnogL6IC8i2OANAwF58AHh0RgQAGwmtRjANxlKiLAFMoILEn6IA1IiELSLIA))

### LINKED ISSUE #617: Crash minifying array destructuring (0.1.4)
Minifying this:
```js
function foo(arr)
{
	let [a, b] = arr;
};
```

Crashes with "SyntaxError: unknown: NodePath has been removed so is read-only. (This is an error on an internal node. Probably an internal error)"

A likely related case is this:

```js
function foo()
{
	return getPromise()
	.then(arr =>
	{
		let [a, b] = arr;
		
		console.log(a);
	});
};
```

Minifying this succeeds, but the destructured let is completely removed, leaving references to nonexistent variables:

```js
function foo() {
	return getPromise().then((b) => {
		console.log(a);  // oops: 'a' declaration removed
	});
}
```

boopathi:
+ Fix #617
+ Fix #615

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -228,21 +228,29 @@ module.exports = ({ types: t, traverse }) => {
               continue;
             }
 
-            if (
-              binding.path.isVariableDeclarator() &&
-              binding.path.node.init &&
-              !scope.isPure(binding.path.node.init) &&
-              binding.path.parentPath.node.declarations
-            ) {
-              if (binding.path.parentPath.node.declarations.length !== 1) {
-                continue;
-              }
-              // Bail out for ArrayPattern and ObjectPattern
+            if (binding.path.isVariableDeclarator()) {
               if (!binding.path.get("id").isIdentifier()) {
+                // deopt for object and array pattern
                 continue;
               }
 
-              binding.path.parentPath.replaceWith(binding.path.node.init);
+              // if declarator has some impure init expression
+              // var x = foo();
+              // => foo();
+              if (
+                binding.path.node.init &&
+                !scope.isPure(binding.path.node.init) &&
+                binding.path.parentPath.node.declarations
+              ) {
+                // binding path has more than one declarations
+                if (binding.path.parentPath.node.declarations.length !== 1) {
+                  continue;
+                }
+                binding.path.parentPath.replaceWith(binding.path.node.init);
+              } else {
+                updateReferences(binding.path, this);
+                removeOrVoid(binding.path);
+              }
             } else {
               updateReferences(binding.path, this);
               removeOrVoid(binding.path);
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -228,21 +228,29 @@ module.exports = ({ types: t, traverse }) => {
...
               }
 
-              binding.path.parentPath.replaceWith(binding.path.node.init);
+              // if declarator has some impure init expression
+              // var x = foo();
+              // => foo();
+              if (
+                binding.path.node.init &&
+                !scope.isPure(binding.path.node.init) &&
+                binding.path.parentPath.node.declarations
```
ebdrup: I'm currently getting the error `Cannot read property 'declarations' of null` here. Doing steps to reproduce is a lot of work (big repo). So I thought I would take a wild stab in the dark and ask if you could see that it would in any way make sense to add a guard here for if `binding.path.parentPath.node` is null? If that doesn't make sense I apologise for wasting your time...

PR #645, opened 2017-08-07
Helper evaluate path

boopathi:


j-f1:
Can you update the README to explain what this does now?

vigneshshanmugam:
path.evaluate in babel doesn't handle some of the cases like this
```
if (v) { var w = 10}
if (w) { console.log(v) } // in diff scope - but still pass isPure check and results in DCE which is wrong
```

The helper takes care of this scenario and also for references used before/after bindings and handles if its within same/diff scope. We need to enable this on simplify and DCE instead of babel evaluate.

packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -8,3 +41,172 @@ module.exports = function evaluate(path) {
     };
   }
 };
+
+// Original Source:
+// https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/evaluation.js
+// modified for Babili use
+function evaluateIdentifier(path) {
+  if (!path.isReferencedIdentifier()) {
```
vigneshshanmugam: check can be removed.

packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -1,4 +1,37 @@
...
+    confident: true
+  };
+
+  // prepare
+  path.traverse({
+    Scope(scopePath) {
+      scopePath.skip();
+    },
+    ReferencedIdentifier(idPath) {
+      const binding = idPath.scope.getBinding(idPath.node.name);
```
vigneshshanmugam: we can remove this as well. evaluate identifier takes care of this already.

packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -8,3 +41,172 @@ module.exports = function evaluate(path) {
...
+    if (!blockParent.get("body").some(stmt => stmt.isAncestor(refPath))) {
+      return { shouldDeopt: true };
+    }
+
+    // Detect usage before init
+    const stmts = fnParent.isProgram()
+      ? fnParent.get("body")
+      : fnParent.get("body").get("body");
+
+    const state = {
```
vigneshshanmugam: lets extract this usage before init in to separate function.

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
```diff
@@ -2472,4 +2454,40 @@ describe("dce-plugin", () => {
...
+    "should optimize to void 0 for lets referenced before init declarations",
+    `
+      function foo() {
+        bar(a);
+        let a = 1;
+      }
+    `,
+    `
+      function foo() {
+        bar(void 0);
```
vigneshshanmugam: should be `referenceError` right not `void 0`

packages/babel-helper-evaluate-path/src/index.js
boopathi: This will help in catching bugs where the refId path is replaced with something else. This has been an issue in DCE+mangle previously.

packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
boopathi: yes. I'll skip this test for now. This doesn't go through `evaluate`.

vigneshshanmugam: Ahh that babel bug? can you put the link to the bug as well?

boopathi: The transformation is not a babel-bug. It's a bug in one-use-replacements in DCE. The parsing is a babel-bug.

vigneshshanmugam: cool

packages/babel-helper-evaluate-path/src/index.js
boopathi: This is required as well. For babel's evaluate to take care of `globals`. Here we ignore globals. In evaluatePath we depot it. If a referencedId makes it to evaluateId, we have to deopt instead of simply ignoring it. Though that code looks like it's never reached, it will be useful for detecting side-effects from other transformations.

vigneshshanmugam:
we need to use this evaluate in simplify as well in patterns matching case, this bug still exists https://github.com/babel/babili/issues/574

boopathi:
Simplify requires a lot of changes in updating scope. Postponing it for later.

vigneshshanmugam:
We are doing the same checks/ optimisations already here https://github.com/babel/babili/blob/master/packages/babel-plugin-minify-dead-code-elimination/src/index.js#L314

boopathi:
Yeah. That code doesn't go through evaluate. It's a one time used variable replacement that doesn't have any side-effects. evaluate takes care of finding values.

PR #649, opened 2017-08-09
Add helper evaluate docs [skip ci]

boopathi:


packages/babel-helper-evaluate-path/README.md
```diff
@@ -1,6 +1,32 @@
 # # babel-helper-evaluate-path
```
existentialism: Btw, just noticed an extra "#" :)

vigneshshanmugam: good catch, addressed https://github.com/babel/babili/commit/391440be53e9cf5a84f138bee5842a998923d42c

PR #658, opened 2017-08-11
Make tdz optional in helper evaluate

boopathi:
For common case, detecting tdz isn't required. Make it optional.

### DIFF
--- a/packages/babel-helper-evaluate-path/src/index.js
+++ b/packages/babel-helper-evaluate-path/src/index.js
@@ -1,6 +1,10 @@
 "use strict";
 
-module.exports = function evaluate(path) {
+module.exports = function evaluate(path, { tdz = false } = {}) {
+  if (!tdz) {
+    return baseEvaluate(path);
+  }
+
   if (path.isReferencedIdentifier()) {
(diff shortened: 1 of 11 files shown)


packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -1,6 +1,10 @@
 "use strict";
 
-module.exports = function evaluate(path) {
+module.exports = function evaluate(path, { tdz = false } = {}) {
+  if (!tdz) {
```
vigneshshanmugam: good catch, would help shave off some time in the benchmarks

packages/babel-preset-minify/src/index.js
```diff
@@ -33,7 +33,8 @@ const PLUGINS = [
 
 const PROXIES = {
   keepFnName: ["mangle", "deadcode"],
-  keepClassName: ["mangle", "deadcode"]
+  keepClassName: ["mangle", "deadcode"],
+  tdz: ["builtIns", "evaluate", "deadcode"]
```
vigneshshanmugam: can you add the same option to preset README?

PR #669, opened 2017-08-16
CLI: reorder files/stdin/help-message

### LINKED ISSUE #666: Running CLI when no TTY
I'm trying to convert my existing `babel` with preset commands on GitLab Pages CI to the new `minify`, but am running into problems because there's no TTY on the Docker runner the CI uses. I find a [related issue](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/2115#note_22926183) which states that 'There is no tty with Gitlab CI'. This means that `process.stdin.isTTY` is always false (see [simple test](https://gitlab.com/probins/testtty/-/jobs/29327772)), and `cli.js` runs `runStdin()` which fails because there are file params.

I can for the moment continue to use `babel`/preset, but I think it would be better if `minify` could handle this situation, perhaps by adding an option to bypass the TTY check if running in this sort of environment?

boopathi:
Fix #666

### DIFF
--- a/packages/babel-minify/src/cli.js
+++ b/packages/babel-minify/src/cli.js
@@ -252,10 +252,12 @@ async function run(args) {
 
   const options = getMinifyOpts(argv);
 
-  if (!process.stdin.isTTY) {
-    return runStdin(argv, options);
-  } else if (argv._.length <= 0) {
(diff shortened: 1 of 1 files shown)


probins:
just to confirm that this does now work on GitLab CI. Thanks!

PR #670, opened 2017-08-16
transform-inline-environment-variables: whitelist option

christophehurpeau:
This PR add a new option:

```json
{
  "plugins": [
    "transform-inline-environment-variables", { "whitelist": ["NODE_ENV", "CI"] }
  ]
}
```

### DIFF
--- a/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
+++ b/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
@@ -1,8 +1,7 @@
 jest.autoMockOff();
 
-const thePlugin = require("../../../utils/test-transform")(
-  require("../src/index")
-);
+const envInliner = require("../src/index");
+const thePlugin = require("../../../utils/test-transform")(envInliner);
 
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
```diff
@@ -21,6 +20,32 @@ describe("inline-env-plugin", () => {
...
+      plugins: [[envInliner, { whitelist: ["NODE_ENV"] }]]
+    }
+  );
+
+  thePlugin(
+    "should not inline environment variables if not present in the whitelist when provided",
+    `
+    process.env.NODE_ENV
+  `,
+    `
```
j-f1: You can remove this second string — the test code will assume that it’s the same.

christophehurpeau: but there is a third argument, should I replace the second by `null`, `undefined` or something else ?

christophehurpeau:
duplicate of #531

PR #672, opened 2017-08-17
transform-inline-environment-variables: include and exclude option

### LINKED ISSUE #513: babel-plugin-transform-inline-environment-variables - Blacklist/whitelist?

christophehurpeau:
Fixes #513
Closes #531 

Same as #531 where test fails

### DIFF
--- a/packages/babel-plugin-transform-inline-environment-variables/README.md
+++ b/packages/babel-plugin-transform-inline-environment-variables/README.md
@@ -30,11 +30,18 @@ npm install babel-plugin-transform-inline-environment-variables
 **.babelrc**
 
 ```json
+// without options
 {
   "plugins": ["transform-inline-environment-variables"]
 }
 ```
 
+```json
+// with options
+{
+  "plugins": ["transform-inline-environment-variables", { "include": ["NODE_ENV", "CI"] }]
+}
+```
 ### Via CLI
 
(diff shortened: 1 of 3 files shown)


boopathi:
Can you change it to `include` and `exclude`.

christophehurpeau:
@boopathi changed as requested

packages/babel-plugin-transform-inline-environment-variables/README.md
```diff
@@ -48,3 +55,8 @@ require("babel-core").transform("code", {
   plugins: ["transform-inline-environment-variables"]
 });
 ```
+
+## Options
+
++ `include` - Array indicating which inline environment variables you want to inline
```
vigneshshanmugam: typo `inline -> include`

christophehurpeau: Thanks

boopathi:
#531 is merged. Thanks.

PR #674, opened 2017-08-17
Improve comments support

### LINKED ISSUE #668: gulp-babili "comments" option has weak default RegExp
For instance, comments like this one from lodash are preserved (pun intended, since that comment includes the string `preserve`):

```/*
     * Creates an array of elements, sorted in ascending order by the results of
     * running each element in a collection thru each iteratee. This method
     * performs a stable sort, that is, it preserves the original sort order of
     * equal elements. The iteratees are invoked with one argument: (value).
     *
     * @static
     * @memberOf _
     * @since 0.1.0
     * @category Collection
     * @param {Array|Object} collection The collection to iterate over.
     * @param {...(Function|Function[])} [iteratees=[_.identity]]
     *  The iteratees to sort by.
     * @returns {Array} Returns the new sorted array.
     * @example
     *
     * var users = [
     *   { 'user': 'fred',   'age': 48 },
     *   { 'user': 'barney', 'age': 36 },
     *   { 'user': 'fred',   'age': 40 },
     *   { 'user': 'barney', 'age': 34 }
     * ];
     *
     * _.sortBy(users, [function(o) { return o.user; }]);
     * // => objects for [['barney', 36], ['barney', 34], ['fred', 48], ['fred', 40]]
     *
     * _.sortBy(users, ['user', 'age']);
     * // => objects for [['barney', 34], ['barney', 36], ['fred', 40], ['fred', 48]]
     */
```

For reference, [this is how other people are doint it](https://github.com/shinnn/uglify-save-license/blob/master/index.js)

_derived from https://github.com/babel/minify/issues/644#issuecomment-322768218_

boopathi:
Fix #668


packages/babel-minify/src/index.js
```diff
@@ -5,26 +5,50 @@ module.exports = function babelMinify(
...
+    ast: false,
+
+    presets: [[minifyPreset, minifyOpts]],
+
     sourceMaps,
-    minified
+    inputSourceMap,
+
+    shouldPrintComment(contents) {
+      return shouldPrintComment(contents, comments);
```
boopathi: This doesn't work right now. Looks like a babel bug ?

vigneshshanmugam: we haven't upgraded babel version for quite sometime right?

xtuc: I believe it's a [babel-generator option](https://github.com/babel/babel/blob/b3372a572db511db44fd151a69375dbb4b8d0573/packages/babel-generator/src/index.js#L43).

boopathi: Yes. It is. I'm not sure why this function is not called at all.

vigneshshanmugam:
Closing in favor of #855

PR #676, opened 2017-08-18
[mangle] blacklist -> exclude

boopathi:
The next version will be `0.3.0`.


Hypnosphi:
I believe it was not a really good idea to update docs before releasing

PR #683, opened 2017-09-01
add transform-hoist-loose-functions plugin

### LINKED ISSUE #682: function not being hoisted in loose mode, causing name collisions
It seems as though `babel-plugin-minify-mangle-names` is not handling loose-mode function scopes correctly:

``` js
// index.js
const foo = (function() {
  const foo = {};
  function outer() { console.log('outer'); }

  if (!process.env.FOO) {
    function inner() { console.log('inner'); }
    function bar() { inner(); }
    foo.inner = bar;
  }

  foo.outer = function() { outer(); };
  return foo;
})();

foo.outer();
```

Running it through `babel-minify` with defaults:

``` sh
$ babel-minify index.js | prettier
```

``` js
// output
const foo = (function() {
  function a() {
    console.log("outer");
  }
  const b = {};
  if (!process.env.FOO) {
    function a() { // <- name clash in loose mode
      console.log("inner");
    }
    b.inner = function() {
      a();
    };
  }
  return (
    (b.outer = function() {
      a();
    }),
    b
  );
})();
foo.outer();
```

For reference, this if the output from `uglify-js`:

```js
var foo = (function() {
  function o() {
    console.log("outer");
  }
  function n() {
    console.log("inner");
  }
  var e = {};
  return (
    process.env.FOO ||
      (e.inner = function() {
        n();
      }),
    (e.outer = function() {
      o();
    }),
    e
  );
})();
foo.outer();
```

rossipedia:
fixes #682

This is my first attempt at a babel plugin (please be gentle 🙂), and this is probably nowhere near the correct way of doing things, but it _seems_ to work (tests included, and existing tests are green). I'm not entirely familiar with the babel API, so there's probably quite a bit of room for improvement. Feedback welcome!

Issues:

* Breaks a bunch of tests if enabled by default, due to issues with `preset-es2015`. IMO this should be opt-in _only_ if you are _not_ using babel to transform ES2015+, as I believe the `transform-es2015-block-scoping` plugin makes this unnecessary.

### DIFF
--- a/packages/babel-plugin-transform-hoist-loose-functions/.npmignore
+++ b/packages/babel-plugin-transform-hoist-loose-functions/.npmignore
@@ -1,0 +1,4 @@
+src
+__tests__
+node_modules
+*.log
--- a/packages/babel-plugin-transform-hoist-loose-functions/README.md
+++ b/packages/babel-plugin-transform-hoist-loose-functions/README.md
@@ -1,0 +1,62 @@
+# babel-plugin-transform-hoist-loose-functions
+
+This plugin hoists functions in loose-mode to their enclosing function or
+program scope.
+
+## Example
+
+**In**
+
+```javascript
+function foo() {
+  if (a) {
+    function bar() {}
+  }
+}
+```
+
+**Out**
+
+```javascript
+function foo() {
+  function bar() {}
+
+  if (a) {}
+}
+```
+
+## Installation
+
+```sh
+npm install babel-plugin-transform-hoist-loose-functions
+```
+
+## Usage
+
+### Via `.babelrc` (Recommended)
+
+**.babelrc**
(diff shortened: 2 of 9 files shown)


rossipedia:
argh.. I'll look into those test failures after lunch

rossipedia:
I'm going to close this PR for now. I've published this plugin on my own ([npm][1] and [github][2]). I'm going to put it through some more paces to make sure it works as I expect.

I'd be happy to submit another PR merging that plugin into this repo to fix #682 if requested.

Thanks!

[1]: https://www.npmjs.com/package/babel-plugin-transform-hoist-loose-functions
[2]: https://github.com/rossipedia/babel-plugin-transform-hoist-loose-functions

Issue #685, opened 2017-09-08
Duplicated names in var statements are considered as different variables


RubenVerborgh:
#### Input Code
```js
function f() {
  var a = 1;
  var b = a;
  var a = 2;
  console.log(a, b);
}
f();
```
or
```js
function f() {
  var a = 1, b = a, a = 2;
  console.log(a, b);
}
f();
```
[Live example on babeljs.io](http://babeljs.io/repl/#?babili=true&browsers=&build=&builtIns=false&code_lz=GYVwdgxgLglg9mABMAFASkQbwFCMQNwEMAnRQxAXkQEYBuXAkxAI0rPr3KoCYPEIEAZzgAbAKYA6EXADmKQgBoWaegF9sqFUA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=react%2Cstage-2%2Cbabili&prettier=false&targets=&version=6.26.0)

#### Actual Output
```js
function f(){var c=1,c=2;console.log(c,c)}f();
```

#### Expected Output

```js
function f(){var c=1,d=c;c=2,console.log(c,d)}f();
```

#### Details
Note the incorrect constant propagation in the `console.log` statement: it has `console.log(c,c)`, even though the two arguments in the original are different. This happens because the variable `a` is declared multiple times as `var`, yet is not recognized as the same variable.

In the correct interpretation `a=2` is recognized as a modification to the same variable, so constant propagation is not allowed to happen. Changing the second `var a` to simply `a` makes the problem disappear.

Slightly more complex example:

```js
function loop() {
  var end = 0;
  var start = end;
  while (end < 10) {
    console.log(start, end);
    var end = end + 1;
  }
}
```

### 2. Stack Trace if there is an Error thrown during minification

None.

### 3. Configuration - babel-minify configuration or babelrc

Default.

The problem is with the `deadcode` option (`babel-plugin-minify-dead-code-elimination`), which is switched on by default.

### 4. Whether you used it with other presets/plugins - like es2015 or env

None.

### 5. Environment: gulp, rollup, webpack, babel-cli, etc...?

All.

PR #686, opened 2017-09-09
Add test for multiple var statements

RubenVerborgh:
Adds a (currently failing) test for #685.

The failure is due to Babel (https://github.com/babel/babel/issues/6217), and passes when a patched Babel (such as https://github.com/babel/babel/pull/6219/) is used.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -406,6 +406,28 @@ describe("dce-plugin", () => {
   );
 
   thePlugin(
+    "should not inline vars modified in different var statements - issue #685",
+    `
+    function f() {
+      var a = 1;
+      var b = a;
+      var a = 2;
+      console.log(a, b);
+    }
+    f();
+  `,
+    `
+    function f() {
+      var a = 1;
+      var b = a;
+      var a = 2;
+      console.log(a, b);
+    }
+    f();
+  `
+  );
+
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
https://github.com/babel/babel/pull/6219 has been merged, so this issue will likely be solved by upgrading to Babel 7.x when it is released.

vigneshshanmugam:
Thanks for the work 👍, We have a existing PR https://github.com/babel/minify/pull/487 which upgrades to babel 7... Will merge this changes once the upgrade is done.

RubenVerborgh:
Pinging since #487 has been merged.

RubenVerborgh:
> Please comment to reopen if you still face the problem.

It's not a matter of having a problem; this PR adds a test that failed incorrectly at the time. Adding this test ensures that this bug cannot creep in anymore.

So yes, please reopen (which I cannot seem to do) and please merge as suggested by @vigneshshanmugam.

boopathi:
Ah. Okay. Issue #685 was fixed by #755 and some tests were added. Do you think the test cases added in #755 would suffice?

Regarding the test cases in this PR: The test setup has changed and the tests are now individual files - https://github.com/babel/minify/blob/master/docs/tests.md#test-files-and-directory-structure.

Can you update this PR to match the current test setup? Thanks.

RubenVerborgh:
Thanks for the pointer. It's indeed covered here: https://github.com/babel/minify/pull/755/files#diff-b8c3b3fafc67f90c48a83ef4f69631aaR11 but actually should already have been addressed by #487, not #755.

PR #610, opened 2017-06-30
Fix folding Array literals (close #608) (close #609)

boopathi:

PR #698, opened 2017-09-25
Fix folding array literals - bail on spread element

### LINKED ISSUE #628: Some of the plugin in this preset delete array reverse call
Babili deletes array reverse call (and probably something else as well)
```bash
npm install babel-preset-babili
echo '[...a].reverse()' | babel --presets=babili
```

And the result is
```js
[...a]
```
`reverse()` disappears for no reasons

### LINKED ISSUE #695: Array spread with function call broken in 0.2.0
#### Input Code

```js
function foo()
{
	return [...bar].reverse();
};
```

#### Actual Output
```js
function foo() {
	return [...bar]; // oops - where did reverse() go?
}
```

#### Expected Output

```js
function foo() {
	return [...bar].reverse();
}
```

#### Details
I'd suspect it is reversing the elements in the array literal and removing the reverse() call, but it does not take in to account spread.
Workaround:
```js
function foo()
{
	const ret = [...bar];
	return ret.reverse();
};
```
This keeps the reverse() call after minify.

boopathi:
+ Fixes #695
+ Fixes #628
+ Related #610

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/src/replacements.js
+++ b/packages/babel-plugin-minify-constant-folding/src/replacements.js
@@ -19,17 +19,24 @@ module.exports = ({ types: t }) => {
     };
   }
 
+  function hasSpread(node) {
+    return node.elements.some(el => t.isSpreadElement(el));
+  }
+
   return {
     ArrayExpression: {
+      canReplace() {
+        return !hasSpread(this.node);
+      },
       members: {
         length() {
-          if (this.node.elements.some(el => t.isSpreadElement(el))) {
+          if (hasSpread(this.node)) {
             return;
           }
           return t.numericLiteral(this.node.elements.length);
         },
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-minify-constant-folding/src/replacements.js
```diff
@@ -51,7 +58,7 @@ module.exports = ({ types: t }) => {
               return evaled.value;
             })
             .join(sep.value);
-          return bad ? undefined : t.stringLiteral(str);
+          return bad ? void 0 : t.stringLiteral(str);
```
j-f1: Why does this use `void 0` instead of `undefined`?

boopathi: we use `void 0` everywhere in the project. Just sticking to that.

Issue #485, opened 2017-03-31
[duplicate] 'let' causing variables to disappear in for-loop


michaelmcleodnz:
This only occurs in conjunction with the es2015 plugin.
The following code
```javascript
function getSum(data) {
  let total = 0;
  for (let i = 0; i < data.length; i++) {
    total += data[i];
  }
  return total;
}
```
gets transpiled to
```javascript
"use strict";function getSum(a){for(var c=0;c<a.length;c++)a[c];return b}
```
which returns the undefined variable `b`. Seemingly `total` has been removed.

The same code using `var i` instead of `let i` produces the expected result
```javascript
"use strict";function getSum(a){for(var c=0,b=0;b<a.length;b++)c+=a[b];return c}
```

hzoo:
Hey @michaelmcleodnz! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

michaelmcleodnz:
If it helps, this seems to be because `var`s declared before a for-loop are pulled inside the for-loop declaration, which can't be done with `let`s because it changes their scope.
Babili knows not to pull the `let`s inside a for-loop without the es2015 plugin, so I'm not sure what the order of operations would be here that causes it to pull the variable inside the for-loop and then decide that it can be optimised away because it is `let`-scoped.

michaelmcleodnz:
I've found that this issue has already been reported in #430 and #412.
[Line 89 of merge-sibling-variables](../blob/b014c3f46aaec5446f914599140c8f259ac35267/packages/babel-plugin-transform-merge-sibling-variables/src/index.js#L89) checks that the for-loop declaration is of type `var`, but presumably somewhere in [minify-dead-code-elimination](../blob/b014c3f46aaec5446f914599140c8f259ac35267/packages/babel-plugin-minify-dead-code-elimination/src/index.js) it is still believed to be of type `let`.

boopathi:
Fixed in #713

armujahid:
This issue is occurring again on latest version of `babel-preset-minify`. For and while loops are eating cont/let declarations immediately before them

[...]

PR #713, opened 2017-11-03
Fix(merge-sibling-var): force recalc ref when concat for-loop var (#485)

### LINKED ISSUE #412: babili made variable lose
Babili 0.0.11

Input: 
```
let tabs = [1,2,3,4,5]
let tabIdx = 0

for (let i = 0; i < tabs.length; i++) {
  // do something
}

console.log({
  tabIdx: tabIdx
})
```

Output:
```
"use strict";
for (var tabs = [1, 2, 3, 4, 5], i = 0; i < tabs.length; i++);
console.log({
  tabIdx: tabIdx
});
```

can not find variable: tabIdx

### LINKED ISSUE #430: Some for loops cause incorrect output
Using `let` in the initialization part of the `for` loop used in the function `foo()` leads to incorrect output.

**Input:**
```js
function foo() {
  let len = 0;
  for (let i = 0; i < 2; i++) {}
  return len;
}
```

**Output:**
```js
"use strict";function foo(){for(var b=0;2>b;b++);return a}
```

The returned variable `a` is not defined. The function `foo()` must always return `0`.

Demo: `https://babeljs.io/repl/#?babili=true&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&code=function foo() {  let len %3D 0%3B  for (let i %3D 0%3B i < 2%3B i%2B%2B) {}  return len%3B}`

### LINKED ISSUE #477: redefine variable when converting let to var
```javascript
{
  let count = 0;
  setInterval(() => {
    console.log('first tick: ', ++count);
  }, 200);
}
{
  let count = 0;
  setInterval(() => {
    console.log('second tick: ', ++count);
  }, 300);
}
```

with babili, will simply converts to

```javascript
{
  var a = 0;
  setInterval(function() {
    console.log('first tick: ', ++a)
  }, 200)
}
{
  var a = 0;
  setInterval(function() {
    console.log('second tick: ', ++a)
  }, 300)
}
```

and getting variable redefined

### LINKED ISSUE #554: Weird mangle error
#### problem

Mangle is incorrect.

#### example

```javascript
// input
function example() {
    let mangleError = [];
    let i = 10;
    while (--i) mangleError.push(i);
    return mangleError;
}

// output
"use strict";function example(){for(var a=[],b=10;--b;)a.push(b);return mangleError}
```

```sh
$ echo \
"function example() {
    let mangleError = [];
    let i = 10;
    while (--i) mangleError.push(i);
    return mangleError;
}" | ./node_modules/.bin/babel --presets=es2015,babili
```

#### package version

- babel-preset-es2015@6.24.1
- babel-preset-babili@0.1.2

---

update:
I found it's caused by `--plugins=minify-simplify,transform-merge-sibling-variables,minify-mangle-names,transform-es2015-block-scoping`

### LINKED ISSUE #581: Babili not validating values properly
If something like this is put through Babili:
```js
let pre = false;
let array = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
let randomNumber = Math.floor(Math.random() * 5);
for(let i = 0; i < array.length; i++)
  if(randomNumber === array[i]){
    pre = true;
    break;
  }
if(!pre)
  console.log("Hello");
```

Then it splits out:
```js
"use strict";for(var array=[1,2,3,4,5,6,7,8,9,10],randomNumber=Math.floor(5*Math.random()),i=0;i<array.length;i++)if(randomNumber===array[i]){break}pre||console.log("Hello");
```
The above mentioned code will fail as the variable `pre` is not defined but is used in a `"strict"` environment. There are other ways to implement the above code like using a `return` instead of `break`. But this methods fails. Looks like a BUG

### LINKED ISSUE #594: Issue with array filled in a loop and returned at the end
Choose one: is this a bug report or feature request?

This used to work with `0.0.12` but as i upgraded to latest `0.1.4` it broke. I am sorry but not sure which plugin might be causing this as i have no idea about the internals of babili.

### Input Code

```js
const getMinQueryNames = (name) => {
    const coll = [];

    for (let i = 0; i < queries.length; i++) {
        const queryName = queries[i].name;

        coll.push(queryName);
        if (queryName === name) {
            break;
        }
    }

    return coll;
};

```

### Babel Configuration (.babelrc, package.json, cli command)

```js
var webpack = require("webpack");
var path = require("path");

module.exports = {
    entry: "./src/index.js",
    output: { 
        filename: "./dist/bundle.js"
    },
    module: {
        rules: [
            {
                test: /\.(js|jsx)$/,
                include: [
                    path.join(__dirname, "src"),
                ],
                use: [
                    {
                        loader: "babel-loader",
                        options: {
                            presets: [["es2015", {"modules": false}], ["babili"]]
                        }
                    }
                ]
            }
        ]
    },
    plugins: [
        new webpack.optimize.ModuleConcatenationPlugin()
    ],
    stats: {
        // Nice colored output
        colors: true
    }
};
```

### Expected Behavior

Output from `0.0.12` version and also the demo site gives this output as well.
Watch the return at end.

```js
var getMinQueryNames=function(a){for(var d,b=[],c=0;c<queries.length&&(d=queries[c].name,b.push(d),d!==a);c++);return b};
```

### Current Behavior

Watch the return at end. At runtime its and error because `coll` is undefined.

```js
var getMinQueryNames=function(a){for(var b,c=[],d=0;d<queries.length&&(b=queries[d].name,c.push(b),b!==a);d++);return coll};
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  6.25
| node             | 8.0
| npm              | 5.0.3
| Operating System | Windows 7 

### UPDATE

I narrowed it down to `mangle` and `mergeVars` plugin. If you use them individually they work fine but if you use them together then they produce this error..

### LINKED ISSUE #720: DCE wrong removing var assignment inside for-loop
This issue is related to es2015 scoping issue and have been fixed on #713.
Now just open for referring in the test case and it should be closed after the PR#713 has been merged. 

#### Input Code

```js
function test () {
  var arr = ['a', 'b', 'c'];
  var len = arr.length;
  var result = null;

  for (let i = 0; i < len; i++) {
      result = arr[i];
  }

  return result || {};
}
```

#### Actual Output
```js
'use strict';function test(){for(var a=['a','b','c'],b=a.length,d=0,;d<b;d++)a[d];return c||{}}
```

#### Expected Output

```js
'use strict';function test(){for(var a=['a','b','c'],b=a.length,d=0,c=0;d<b;d++)c=a[d];return c||{}}
```

garyyeap:
Fix #594
Fix #430
Fix #412 
Fix #554
Fix #581
Fix #477 
Fix #720

### DIFF
--- a/packages/babel-plugin-transform-merge-sibling-variables/src/index.js
+++ b/packages/babel-plugin-transform-merge-sibling-variables/src/index.js
@@ -90,10 +90,15 @@ module.exports = function({ types: t }) {
               return;
             }
 
-            init.node.declarations = node.declarations.concat(
+            const declarations = node.declarations.concat(
               init.node.declarations
             );
+
+            // temporary workaround to forces babel recalculate scope,
+            // references and binding until babel/babel#4818 resolved
             path.remove();
+            init.remove();
+            next.node.init = t.variableDeclaration("var", declarations);
           }
         ]
       }
--- a/packages/babel-plugin-transform-merge-sibling-variables/src/index.js
+++ b/packages/babel-plugin-transform-merge-sibling-variables/src/index.js
@@ -97,8 +97,7 @@ module.exports = function({ types: t }) {
             // temporary workaround to forces babel recalculate scope,
             // references and binding until babel/babel#4818 resolved
             path.remove();
-            init.remove();
-            next.node.init = t.variableDeclaration("var", declarations);
+            init.replaceWith(t.variableDeclaration("var", declarations));
           }
         ]
       }
--- a/packages/babel-preset-minify/__tests__/babili-es2015-tests.js
+++ b/packages/babel-preset-minify/__tests__/babili-es2015-tests.js
@@ -64,4 +64,111 @@ describe("preset along with es2015", () => {
       }
     `
   );
+
+  thePlugin(
+    "should fix issue#485",
+    `
+      function getSum(data) {
+        let total = 0;
+        for (let i = 0; i < data.length; i++) {
+          total += data[i];
+        }
+        return total;
+      }
+    `,
+    `
+      "use strict";
+
+      function getSum(a) {
+        for (var b = 0, c = 0; c < a.length; c++) b += a[c];
+        return b;
+      }
+      `
+  );
+
+  thePlugin(
+    "should fix issue#412",
+    `
+      let tabs = [1,2,3,4,5];
+      let tabIdx = 0;
+
+      for (let i = 0; i < tabs.length; i++) {
+        // do something
+      }
+
+      console.log({
+        tabIdx: tabIdx
+      });
+    `,
+    `
+      "use strict";
+
+      for (var tabs = [1, 2, 3, 4, 5], tabIdx = 0, i = 0; i < tabs.length; i++);
+
+      console.log({
+        tabIdx: tabIdx
+      });
+    `
+  );
+
+  thePlugin(
+    "should fix issue#477",
+    `
+      {
+        let count = 0;
+        setInterval(() => {
+          console.log('first tick: ', ++count);
+        }, 200);
+      }
+      {
+        let count = 0;
+        setInterval(() => {
+          console.log('second tick: ', ++count);
+        }, 300);
+      }
+    `,
+    `
+      'use strict';
+
+      {
+        var count = 0;
+        setInterval(function () {
+          console.log('first tick: ', ++count);
+        }, 200);
+      }
+      {
+        var _count = 0;
+        setInterval(function () {
+          console.log('second tick: ', ++_count);
+        }, 300);
+      }
+    `
+  );
+
+  thePlugin(
+    "should fix issue#720",
+    `
+      function test () {
+        var arr = ['a', 'b', 'c'];
+        var len = arr.length;
+        var result = null;
+
+        for (let i = 0; i < len; i++) {
+            result = arr[i];
+        }
+
+        return result || {};
+      }
+    `,
+    `
+      'use strict';
+
+      function test() {
+
+        for (var a = ['a', 'b', 'c'], b = a.length, c = null, d = 0; d < b; d++) c = a[d];
+
+        return c || {};
+      }
+    `
+  );
 });


garyyeap:
Not sure how to write test as some issues only occur when using mergeVars, deadcode, mangle at the same time

boopathi:
Thanks a lot. I'll merge it with tests. You can add test cases in `babel-preset-minify/__tests__`

packages/babel-plugin-transform-merge-sibling-variables/src/index.js
```diff
@@ -90,10 +90,15 @@ module.exports = function({ types: t }) {
...
-            init.node.declarations = node.declarations.concat(
+            const declarations = node.declarations.concat(
               init.node.declarations
             );
+
+            // temporary workaround to forces babel recalculate scope,
+            // references and binding until babel/babel#4818 resolved
             path.remove();
+            init.remove();
+            next.node.init = t.variableDeclaration("var", declarations);
```
boopathi:
```js
next.replaceWith(t.variableDeclaration("var", declarations))
```

boopathi: Can you also fix this one? Use babel path API instead of mutating node.

garyyeap: Tried & failed with this way, the output eliminated the for-loop

boopathi: Oops. It's 

```js
init.replaceWith(t.variableDeclaration("var", declarations);
```

and `init.remove()` should be removed.

garyyeap: Done, thanks for correcting

edmorley:
Hi!

This is breaking the build in hueniverse/hawk#223 - I don't suppose a point release could be made that includes this fix? :-)

boopathi:
@edmorley Every commit to master is released as a `canary` version - docs - https://github.com/babel/minify/tree/master/docs#canary-version ... But it's ONLY for ease of testing. Would not recommend using `canary` tag committed in your package.json.

allanjackson:
I just want to chime in that this is also breaking our build, so we can't use babel-minify 0.2.0 right now. It would be great if we could get a non-canary release that includes this fix.

vigneshshanmugam:
@allanjackson Its done. use [0.3.0](https://github.com/babel/minify/releases/tag/gulp-babel-minify%400.3.0)

allanjackson:
Thanks!

PR #714, opened 2017-11-07
Specify *.jar as binary in gitattributes

remcohaszing:
scripts/gcc.jar always showed up as being modified, because git was
replacing the newlines in this binary file.


boopathi:
Thanks!

PR #717, opened 2017-11-08
Fixtures

boopathi:
Move single file to autoupdating `actual.js` and `expected.js` - similar to babel/babel repository - with jest.

This is helpful to upgrade to babel-7 - simply by updating all the fixtures.


j-f1:
How can I enable `--update-fixtures` when running `yarn test`?

boopathi:
Unfortunately I haven't figured out a way to support that. Right now using jest CLI for a particular package works. 

```sh
jest packages/babel-plugin-minify-mangle-names -- --update-fixtures
```

Also, to update all fixtures

```sh
jest ./ -- --update-fixtures
# the ./ seems to be relevant. Otherwise jest treats --update-fixtures
# as directory for some weird reason
```

PR #722, opened 2017-11-13
refactor(tests): Tests to fixtures - typeConstructors plugin

boopathi:


packages/babel-plugin-minify-type-constructors/__tests__/fixtures/basic/expected.js
```diff
@@ -0,0 +1,5 @@
+!!x;
++x;
+x + "";
```
j-f1: Shouldn’t this be `x+""` (i.e. no spaces around the `+`)?

```js
x+""
```

boopathi: This is taken care of by babel-generator's `minify:true` mode.

PR #725, opened 2017-11-14
Fix for dead code elimination trying to place an expression into a JSXIdentifier.

### LINKED ISSUE #724: dead-code-elimination erroneously tries to inline expressions into JSXIdentifiers.
Reproduction on RunKit: https://runkit.com/tolmasky/dead-code-inlines-jsxidentifiers

```javascript
function Input(element)
{
    var ElementType = element || "input";

    return <ElementType ref = "input" />;
}
```

This will try to put `(element || "input")` into the actual JSX element, (something like `<(element||"input") ref = "input/>`), which is not allowed, and will thus throw this error:

TypeError: unknown: Property name of JSXOpeningElement expected node to be of a type ["JSXIdentifier","JSXMemberExpression"] but instead got "LogicalExpression"

It is important that I run this before my actual React transforms, since there are in-between transforms that rely on the right code having been removed, so I can not use as a temporary workaround switching the order of the transformations.

tolmasky:
Fix for dead code elimination trying to place an expression into a JSXIdentifier. Closes #724.

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/dead-code-elimination-test.js
@@ -200,4 +200,25 @@ describe("dce-plugin", () => {
       plugins: [[deadcode, { tdz: true }]]
     }
   );
+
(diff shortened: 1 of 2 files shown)


boopathi:
Commented in #724

PR #736, opened 2017-11-19
Optimize if-statement tests

j-f1:
These optimizations are only safe when applied to expressions where only the truthiness or falsiness​ of the result is important. Therefore, I’ve applied them to:

1. Ternary tests
2. `if` or `while` statement tests
3. The LHS of logical operators that are either:
   1. the only expression in an `ExpressionStatement` or
   2. not the last expression in a `SequenceExpression`

Ref #25.


j-f1:
@boopathi This should be good to go.

PR #759, opened 2017-12-16
refactor(tests) Tests to fixtures - inline env

vigneshshanmugam:


boopathi:
The test there tests for `true` output. And we should not try to fit this use case to fixtures. Let it be in the single file only.

boopathi:
Closing this as fixtures not required for this plugin.

PR #760, opened 2017-12-16
refactor(tests) Tests to fixtures - propertyLiterals

vigneshshanmugam:


packages/babel-plugin-transform-property-literals/__tests__/fixtures/invalid-identifiers/actual.js
```diff
@@ -0,0 +1,7 @@
+// FIXME: The test name states that no transformation should take place,
+// but the actual test specifies that the quotes should be stripped.
+//prettier-ignore
+({
+  "default": null,
```
vigneshshanmugam: Isn't this a safe transformation?. We can remove the FIXME from here AFAIK.

packages/babel-plugin-transform-property-literals/__tests__/fixtures/invalid-es5-property/expected.js
```diff
@@ -0,0 +1,5 @@
+// prettier-ignore
+({
+  "℘": "wp",
```
vigneshshanmugam: jest transforms it while updating fixtures.

PR #761, opened 2017-12-19
babel 7 compatibility change - mangle names: 

loborobo:
handle getFunctionParen not returning a Program

Due to [this](https://github.com/babel/babel/pull/5923) change, mangle-names does not work with Babel 7 without this update.


boopathi:
Thanks for the PR.

Looks like the lint failed. You can fix it by running `yarn fix`

loborobo:
np. 😄

PR #762, opened 2017-12-19
refactor(tests) tests to fixtures - regexpConstructors

vigneshshanmugam:


packages/babel-plugin-transform-regexp-constructors/__tests__/fixtures/escape/actual.js
```diff
@@ -0,0 +1 @@
+String.raw`var x = new RegExp('\\w+\\s')`;
```
boopathi: The string.raw should be removed

vigneshshanmugam: ahh missed it. will update.

Issue #766, opened 2017-12-19
Drop Node 4?


boopathi:
+ Jest 22 stops supporting node 4 - http://facebook.github.io/jest/blog/2017/12/18/jest-22.html
+ webpack drops node 4 support - eg. https://github.com/webpack-contrib/babel-minify-webpack-plugin/pull/70

xtuc:
I think you'll be forced to since Babel dropped Node 4.

boopathi:
Ah. yes. That makes it easier. Thanks. Closing this then. 

Decision: dropping node 4

PR #767, opened 2017-12-20
Remove node 4. Add node 9

boopathi:
Ref: #766 - decision was made here.

Late to the game?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,7 +4,7 @@
       "env",
       {
         "targets": {
-          "node": 4
+          "node": 6
         },
         "include": ["transform-async-to-generator"],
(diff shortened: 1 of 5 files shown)


azz:
Why is everyone dropping node 4 support so soon? I know LTS is ending next year,  but still seems early?

boopathi:
@azz I'm not sure either.

xtuc:
I was incredibly wrong, **Babel didn't dropped Node 4**.

> Node.js 0.10 and 0.12 has been dropped

azz:
Babel dropped node v4 on the root package.json (i.e. just for development).

PR #773, opened 2017-12-29
(perf) - move the collapser check before inling properties

vigneshshanmugam:
This yields us a huge **3.5x improvement** on a test file > 2MB. 
```
pluginAlias                       time(ms)

// Before
transform-inline-consecutive-adds - 5716.129
transform-inline-consecutive-adds - 5752.153
transform-inline-consecutive-adds - 5813.064

// After
transform-inline-consecutive-adds - 1624.955
transform-inline-consecutive-adds - 1925.298
transform-inline-consecutive-adds - 1788.473
```
Thanks to chrome devtools for helping to find the root cause. Confirmation from devtools

### Before

screen shot 2017-12-29 at 22 16 38

### After

screen shot 2017-12-29 at 22 27 23

### DIFF
--- a/packages/babel-plugin-transform-inline-consecutive-adds/src/index.js
+++ b/packages/babel-plugin-transform-inline-consecutive-adds/src/index.js
@@ -149,11 +149,11 @@ function getReferenceChecker(references) {
 function tryUseCollapser(t, collapser, varDecl, topLevel, checkReference) {
   // Returns true iff successfully used the collapser. Otherwise returns undefined.
   const [name, init, startIndex] = topLevel;
-  const body = varDecl.parentPath.get("body");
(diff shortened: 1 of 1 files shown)


sophiebits:
.get('body') creating an array seems like a huge performance trap @hzoo. Same thing I ran into in https://github.com/babel/babel/pull/7028.

vigneshshanmugam:
Thanks @sophiebits for pointing out. Good to know. I am also seeing smaller perf slowdown on get calls from path not just on 'body'. It happens on all array elements

Probably it has something to do with https://github.com/babel/babel/blob/068c65a1dd5bf606d0441506f2f5773be168223b/packages/babel-traverse/src/path/family.js#L104

sophiebits:
Yes.

PR #777, opened 2018-01-14
Add Node 4 back

boopathi:
Related: https://github.com/babel/minify/pull/767#issuecomment-353271587 and #766


boopathi:
Ah. Jest doesn't support node-4!

xtuc:
Thanks, sorry for the wrong call.

PR #778, opened 2018-01-15
(perf) - Improve performance of simplify by removing traversal

vigneshshanmugam:
This PR handles the type information that is lost due to sequence expression transformation in Simplify plugin. It takes a different approach instead of traversing all of the identifiers for type information. 

### Before

pluginAlias                             time(ms) # visits time/visit(ms)
// Run 1
minify-simplify                         6090.013 278330   0.022
transform-simplify-comparison-operators 41.059   18348    0.002
// Run 2
minify-simplify                         6029.847 278330   0.022
transform-simplify-comparison-operators 46.415   18348    0.003
// Run 3
minify-simplify                         6045.218 278330   0.022
transform-simplify-comparison-operators 44.464   18348    0.002

### After
// 1
minify-simplify                         4617.782 278330   0.017
transform-simplify-comparison-operators 122.181  18348    0.007
// 2
minify-simplify                         4718.766 278330   0.017
transform-simplify-comparison-operators 122.642  18348    0.007
// 3
minify-simplify                         4447.601 278330   0.016
transform-simplify-comparison-operators 127.682  18348    0.007

The type inference is based on the Babel inference. 

+ Only for identifiers we take the different path

### DIFF
--- a/packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
+++ b/packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
@@ -1,6 +1,67 @@
 "use strict";
 
-module.exports = function() {
+module.exports = function({ types: t }) {
+  // custom implementation of getTypeAnnotation that fixes
+  // the type information that is lost during sequence expression transformation
+  // https://github.com/babel/minify/issues/323
+  function customTypeAnnotation(path) {
+    if (path.typeAnnotation) {
+      return path.typeAnnotation;
+    }
+
+    // We are not handling the case of literals and other base types
+    // since they are already handled via getTypeAnnotation
+    const { node } = path;
+    const binding = path.parentPath.scope.getBinding(node.name);
+
+    const types = [];
+    if (binding && binding.constantViolations) {
+      if (binding.identifier.typeAnnotation) {
+        return binding.identifier.typeAnnotation;
+      }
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-transform-simplify-comparison-operators/src/index.js
```diff
@@ -1,6 +1,67 @@
...
+
+    leftTypes = [].concat(leftTypes, left.getTypeAnnotation());
+    rightTypes = [].concat(rightTypes, right.getTypeAnnotation());
+
+    leftTypes = t.createUnionTypeAnnotation(leftTypes);
+    rightTypes = t.createUnionTypeAnnotation(rightTypes);
+
+    if (
+      !t.isAnyTypeAnnotation(leftTypes) &&
+      t.isFlowBaseAnnotation(leftTypes)
```
boopathi: I'm not sure if this affects the babel-7 upgrade. I think we should test this against Babel7. What do you think?

vigneshshanmugam: No it stays the same, No change for babel 7 

https://github.com/babel/babel/blob/64dfb7574e924be3ad737c33cbc81d619dbc14c6/packages/babel-traverse/src/path/inference/index.js#L109-L111

PR #786, opened 2018-01-29
Replaced gulp-util dependency with fancy-log.

FizzyGalacticus:
The gulp-util package is being officially deprecated in Gulp v4, and has already started causing issues for my team.

Following this article: https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5

...I have removed it in favor of fancy-log, which seems to be just the logging utilitiy from gulp-util anyways.

Running `yarn` and `yarn build` seemed to be unaffected, and no errors were present., therefore I believe this to be an adequate fix for this problem.


package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "gulp-babel-minify",
+  "version": "0.2.0",
```
boopathi: these are not required

FizzyGalacticus:
I removed the unnecessary commits to the package.json file.

package.json
FizzyGalacticus: I removed the unnecessary commits to the package.json file.

PR #789, opened 2018-02-02
Use hexadecimal notation for number literals when shorter

maurobringolf:
According to https://stackoverflow.com/a/9459210/6374865 there are integers whose hexadecimal representation is shorter than decimal (and exponential). I asked on Slack if this is a valid optimization, but realized that it is not much code to write: Compute hexadecimal representation with `toString(16)` and compare lengths. However, I think it is only valid for positive integers so I added a check for that.

Is this a valid idea/optimization?

*There is one test with an illustrating example where one byte is saved.*


packages/babel-plugin-minify-numeric-literals/src/index.js
```diff
@@ -25,9 +26,13 @@ module.exports = function() {
             .replace(/e0/, "");
         }
 
-        const replacement =
+        let replacement =
           normal.length > exponential.length ? exponential : normal;
 
+        if (path.node.value > 0 && Number.isInteger(path.node.value))
```
vigneshshanmugam: small micro optimisation we can do is to compute the hex only for positive integers. Move this block
```js
const hexadecimal = `0x${path.node.value.toString(16)}`;
```
inside the if statement.

maurobringolf: Thanks, I moved it int 9fe2436. I think this can be made to work for the same integers with opposite signs but haven't verified that yet.

taion:
For a file full of vanilla numbers, this increases file size after gzipping.

If I generate 3 files full of random 64-bit unsigned integers, in respectively decimal, hex without `0x`, and hex with `0x`, I see that the file with decimals is largest:

```
$ ll
total 10976
-rw-r--r--  1 jjia  staff  2035712 Feb  3 16:32 dec.txt
-rw-r--r--  1 jjia  staff  1691648 Feb  3 16:32 hex.txt
-rw-r--r--  1 jjia  staff  1892352 Feb  3 16:32 hex2.txt
```

However, after gzipping, it's the smallest:
```
$ ll
total 5840
-rw-r--r--  1 jjia  staff   977581 Feb  3 16:32 dec.txt.gz
-rw-r--r--  1 jjia  staff   980171 Feb  3 16:32 hex.txt.gz
-rw-r--r--  1 jjia  staff  1027795 Feb  3 16:32 hex2.txt.gz
```

Hard to say what this means for a codebase, but this is not presumptively an improvement in scenarios with compression.

j-f1:
What happens if you convert all of the numbers to binary (`0b1011`)? That might decrease the gzip size. If so, maybe we should have a `gzip` option. When it’s false, we attempt to produce the shortest output possible. When it’s `true`, we try to maximize repetition.

taion:
That's unlikely to help. Ultimately the same amount of information in the digits proper doesn't change, so net you end up just paying the cost of the unnecessary `0b` everywhere.

maurobringolf:
Have there been other cases like this before? Is gzip the main target to be optimized for? Then this PR should not be merged I guess.

boopathi:
Thanks for trying it out.

Gzip is the main target and we display these numbers in the README consider the gzip win as a comparison to other minifiers. Please comment to reopen this if things change later.

PR #807, opened 2018-02-28
fix: Remove unnecessary checks for constant folding

diervo:
We found that `traverse.hasType` introduces a lot of side-effects since its create cache entries for path where scope might not be defined.

If this method is called for a path that has never been created, it will bound the wrong scope object to it, causing subsequent access to throw.

For example this exception:
```
TypeError: Cannot read property 'getBinding' of null
```
can be traced down to this exact problem.
FYI this is the exact same failure we have on the CI.

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/src/index.js
+++ b/packages/babel-plugin-minify-constant-folding/src/index.js
@@ -116,10 +116,6 @@ module.exports = babel => {
           return;
         }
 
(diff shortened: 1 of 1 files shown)


boopathi:
Good find! Thanks.

diervo:
@boopathi It took an international flight of debugging to figure out xDD

PR #814, opened 2018-04-02
Add member access constant folding

fathyb:
Simplifies member access when it's safe (where `!($key in {})` and defined after any spread or computed property).

In : 
```js
const foo = {b: 'ar'}.b
```

Out : 
```js
const foo = 'ar'
```

- [x] bailout on computed properties or object spread
- [x] shorthand properties (`{foo}.foo` => `foo`)
- [x] numeric literals (`{0:'foo'}[0]` => `'foo'`)

Sorry for the after-review commits, just noticed it lacked numeric literals support

### DIFF
--- a/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
@@ -1,0 +1,27 @@
+({
+  foo: 'oof',
+  bar: 'rab'
+}).foo;
+({
+  bar: 'rab'
+}).foo;
+({
+  bar: 'rab'
+}).toString;
+({
+  foo,
+  bar: 'test',
+  baz
+}).foo;
+({
+  bar: 'test',
+  baz
+}).foo;
+({
+  0: 'foo',
+  1: 'bar',
+  b: 'az'
+})[0];
+({
+  foo: 'bar'
+})[0];
--- a/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/expected.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/expected.js
@@ -1,0 +1,7 @@
+'oof';
+void 0;
+({}).toString;
+foo;
+void 0;
+'foo';
+void 0;
--- a/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access-builtin/actual.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access-builtin/actual.js
@@ -1,0 +1,16 @@
+({
+  toString: () => 'bar',
+  [computed]: 'foo'
+}).toString;
+({
+  [computed]: 'foo',
+  toString: () => 'bar'
+}).toString;
+({
+  [computed]: 'foo'
+}).toString;
+({
+  toString: () => 'bar'
+}).toString;
+({}).toString;
+({}).notBuiltIn;
--- a/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access-builtin/expected.js
+++ b/packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access-builtin/expected.js
@@ -1,0 +1,15 @@
+({
+  toString: () => 'bar',
+  [computed]: 'foo'
+}).toString;
+
+() => 'bar';
+
+({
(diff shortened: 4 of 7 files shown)


packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
```diff
@@ -0,0 +1,13 @@
...
+({['foo']: 'oof', bar: 'rab'}).baz;
+({bar: 'rab'}).toString;
+({
+  foo: 'bar',
+  [Math.random()]: 'foo'
+}).foo;
+({
+  [Math.random()]: 'foo',
+  foo: 'bar'
+}).foo;
```
xtuc: Could we also test for:

```js
({ foo: fn() }).foo
```

That shouldn't be inlined since `fn()` could cause side effects.

packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
```diff
@@ -0,0 +1,13 @@
+({foo: 'oof', bar: 'rab'}).foo;
+({bar: 'rab'}).foo;
+({['foo']: 'oof', bar: 'rab'}).foo;
+({['foo']: 'oof', bar: 'rab'}).baz;
+({bar: 'rab'}).toString;
```
xtuc: It's an edge but this one can be unsafe. Nothing prevents the user from overriding `Object.prototype.toString`.

packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
fathyb: It seems that we do the same for arrays : 
`[0, sideEffect()][0]` gets inlined to `0`.

If you still think we should do it I think we can bailout if we recursively find a call expression in the initialiser and maybe share the logic with the array replacer.

xtuc: Ok I didn't know about that. I haven't heard about issues with the array replacer. To me that's safe enough, but a user has no way to know/debug that.

I'm wondering if we shouldn't introduce flags, like GCC's `-O1` to `-Os`.

packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
fathyb: Good catch, I pushed a commit with a fix and more tests

packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/actual.js
fathyb: I second both, I think we should warn the user by default and then use multiple optimization level options to silent the warnings and choose if we should assume not side effects.

We could also offer the user to annotate its calls using comments, [à la Closure](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler). Something like : 

```js

/** @babel:side-effect **/
function foo() {
  sideEffectApi()
}

/** @babel:side-effect:expression:bar() **/
const dontTrim = bar()
```

The is also the [`sideEffects` proposal](https://github.com/webpack/webpack/tree/master/examples/side-effects) from Webpack which may be handy, although I think we're not module aware.

I feel like this needs its own discussion, perhaps I should open an issue?

vigneshshanmugam: Yeah totally agree, lets create a separate issue and discuss different methods there.

boopathi:
nit: Can you split the test cases into multiple files?

packages/babel-plugin-minify-constant-folding/__tests__/fixtures/member-access/expected.js
```diff
@@ -0,0 +1,7 @@
+'oof';
+void 0;
+({}).toString;
```
j-f1: This is unsafe — what would happen if `toString` was overridden to call `JSON.stringify`?

xtuc: https://github.com/babel/minify/pull/814#discussion_r178538385.

I really think we should clear https://github.com/babel/minify/issues/815 first.

PR #831, opened 2018-05-03
fix: account for different scopes during path evaluation

### LINKED ISSUE #574: if statement wrongly removed when var declaration inside conditional block

### LINKED ISSUE #810: "if" condition is incorrectly omitted after minify
#### Input Code

```js
if (false) {
  var bar = true;
}
if (bar) {
  alert('bug!');
}
```

#### Actual Output

```js
var bar;alert('bug!');
```

#### Expected Output

```js
var bar;bar&&alert('bug!');
```

#### Details

The if condition is incorrectly removed. The logic is changed.

I know that putting var inside a block scope is definitely a bad practice. I've just tried to minify some legacy codes and encountered this critical issue.

Example: [See it here in repl](https://babeljs.io/repl#?babili=true&browsers=&build=&builtIns=false&code_lz=JYMwBAFCCGA2DOBTAlGA3gKDGAbtATmAEYFgC8YALvgK6IDcGAvhqJCfqptnIvpRADkRGgHMAhIOSMWGIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=babili&prettier=false&targets=&version=6.26.0&envVersion=)

vigneshshanmugam:
+ fix #810 

+ fix #574  properly

### DIFF
--- a/packages/babel-helper-evaluate-path/src/index.js
+++ b/packages/babel-helper-evaluate-path/src/index.js
@@ -1,12 +1,14 @@
 "use strict";
 
+const t = require("@babel/types");
+
 module.exports = function evaluate(path, { tdz = false } = {}) {
-  if (path.isReferencedIdentifier()) {
-    return evaluateIdentifier(path);
+  if (!tdz && !path.isReferencedIdentifier()) {
+    return baseEvaluate(path);
   }
 
-  if (!tdz) {
-    return baseEvaluate(path);
+  if (path.isReferencedIdentifier()) {
+    return evaluateIdentifier(path);
   }
 
   const state = {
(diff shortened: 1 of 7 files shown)


packages/babel-plugin-minify-dead-code-elimination/src/index.js
```diff
@@ -976,10 +976,14 @@ module.exports = ({ types: t, traverse }) => {
     } else {
       path.traverse({
         VariableDeclaration(varPath) {
-          if (!varPath.isVariableDeclaration({ kind: "var" })) return;
+          const { node } = varPath;
+
+          if (node.kind !== "var") {
+            return;
+          }
```
boopathi: why is this changed? Let's stick to `.isType( params )`

boopathi: isVariableDeclaration has `kind` ... So let's change it back.

packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -113,8 +113,12 @@ function evaluateBasedOnControlFlow(binding, refPath) {
       return { shouldDeopt: true };
     }
 
-    let blockParent = binding.path.scope.getBlockParent().path;
     const fnParent = binding.path.getFunctionParent();
+    if (!fnParent) {
+      return { shouldDeopt: true };
+    }
+
+    let blockParent = binding.path.scope.getBlockParent().path;
```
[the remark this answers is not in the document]
vigneshshanmugam: Block parent cannot be null, it returns Program path.

PR #834, opened 2018-05-08
fix(deadcode): account for impure paths inside for statements

### LINKED ISSUE #824: simplify's loop rewriting incorrectly removes preceding variable declarations
#### Input Code

```js
let foo;
while (0) {}
console.log(foo);
```

#### Actual Output

```js
'use strict';console.log(foo);
```

#### Details

Minimal repro in the sandbox [here](https://babeljs.io/repl/#?babili=true&browsers=&build=&builtIns=false&code_lz=DYUwLgBAZg9jEF4IHJY2QbgFAHcAWAlqBABQAMAlBAN4C-WAxjAHYDOMoAdMDAOYloKGIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Cstage-2%2Cstage-3%2Cbabili&prettier=false&targets=&version=6.26.0&envVersion=).

Passing `simplify: false` to `babel-preset-minify` resolves the issue.

It looks like what's happening is that the loop is rewritten to a for loop, then preceding statements are sometimes moved into the initialization segment of the for loop. But if the initialization segment is empty, or if the entire for loop is removed, then the declarations are lost. 

Note that this does not require the loop to be completely deleted, and that more than one preceding statement can be lost. Our original code that prompted this looks roughly like:

[Sandbox](https://babeljs.io/repl/#?babili=true&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAZiEMC8MDkCRoNwChSSwBGAhgE4rqlk64Bu5MpAXpQI4CuAHgBQCUeAO4ALAJYAbAKY8WfGAG9cMJiVapOvAbgC--cBBBSAdOJABzHphgBqFWS1A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Cstage-2%2Cstage-3%2Cbabili&prettier=false&targets=&version=6.26.0&envVersion=)
```js
const foo = 'foo';
const bar = 'bar';
let baz = qux();
while(baz) {
  baz = qux();
}
console.log(foo + bar);
```

And we're seeing the declarations for both foo and bar deleted.

vigneshshanmugam:
+ fix #824

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/loops/actual.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/loops/actual.js
@@ -19,6 +19,10 @@ for (; true; ) {
 for (; false; ) {
   baz();
 }
+for (var j; false; ) {
+  blah();
+}
+for (0; false; ) {}
 
 // do_while
 do {
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/loops/expected.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/loops/expected.js
@@ -16,6 +16,8 @@ for (;;) {
   bar();
 }
 
+var j;
+
 // do_while
 do {
   foo();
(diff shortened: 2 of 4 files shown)


j-f1:
What happens if there’s a side-effectful expression in the update/condition parts of the loop?

vigneshshanmugam:
Shouldn't be a problem since we are doing this transform only if test statement is pure. The update path won't be executed at all in these cases.

Issue #839, opened 2018-05-13
minify-infinity not doing anything


Cyp:
**Describe the bug**

The minify-infinity pass doesn't do anything.

**To Reproduce**

```js
const infinities = [Infinity, 1/0, true/false, !0/!1, true/(true/(true/false))];
const zeroes = [1/Infinity, 1/(1/0), false/true, !1/!0, true/(true/false)];
const justone = Infinity;
alert(Infinity);
```

**Actual Output**

```js
"use strict";var infinities=[Infinity,1/0,!0/!1,!0/!1,!0/0],zeroes=[1/Infinity,0,0,0,0],justone=Infinity;alert(Infinity);
```

**Expected Output**

```js
"use strict";var infinities=[1/0,1/0,1/0,1/0,1/0],zeroes=[0,0,0,0,0],justone=1/0;alert(1/0);
```

**Additional context**

It returns here, and therefore doesn't do the substitution:
```js
        if (path.isLVal() && !path.parentPath.isExpressionStatement()) {
          return;
        }
```
The `path.parentPath.type` is `"ArrayExpression"`, `"BinaryExpression"`, `"VariableDeclarator"` and `"CallExpression"`, respectively.

I couldn't find any documentation on what `.isExpressionStatement()` is actually supposed to do, so I don't know whether the bug is in minify-infinity or in babel itself.

boopathi:
Looks like this regressed in #241 ...

PR #840, opened 2018-05-14
fix: Make babel-plugin-minify-infinity work again. (#839)

Cyp:
Here's an attempt at fixing #839. I don't know whether there are any other cases that need catching.

### DIFF
--- a/packages/babel-plugin-minify-infinity/src/index.js
+++ b/packages/babel-plugin-minify-infinity/src/index.js
@@ -6,11 +6,18 @@ module.exports = function({ types: t }) {
     t.numericLiteral(1),
     t.numericLiteral(0)
   );
+  const badTransforms = {
+    ArrayPattern: () => true,
+    AssignmentExpression: path => path.container.left === path.node,
+    ObjectProperty: path => path.container.shorthand,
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-minify-infinity/src/index.js
```diff
@@ -6,6 +6,12 @@ module.exports = function({ types: t }) {
     t.numericLiteral(1),
     t.numericLiteral(0)
   );
+  const badTransforms = {
+    ArrayPattern: true,
+    AssignmentExpression: true,
```
boopathi: Doesn't this prevent transforming `x = Infinity`?

boopathi:
Ah!. I worked on the same thing and have #841 .

packages/babel-plugin-minify-infinity/src/index.js
Cyp: Yes, I missed testing that case.

Cyp:
Don't know if it's the right approach, but I've updated my version in case.

vigneshshanmugam:
Thanks for the PR @Cyp

Fixed by #841

PR #841, opened 2018-05-14
fix: minify infinity in expressions

### LINKED ISSUE #839: minify-infinity not doing anything

boopathi:
Fix #839

### DIFF
--- a/packages/babel-plugin-minify-infinity/__tests__/fixtures/expressions/expected.js
+++ b/packages/babel-plugin-minify-infinity/__tests__/fixtures/expressions/expected.js
@@ -1,6 +1,6 @@
 let x = [1 / 0, 1 / 0];
 let y = [{
-  a: Infinity
+  a: 1 / 0
 }, {
-  a: Infinity
+  a: 1 / 0
 }];
--- a/packages/babel-plugin-minify-infinity/__tests__/fixtures/expressions/expected.js
+++ b/packages/babel-plugin-minify-infinity/__tests__/fixtures/expressions/expected.js
@@ -1,0 +1,6 @@
+let x = [1 / 0, 1 / 0];
+let y = [{
+  a: Infinity
+}, {
+  a: Infinity
+}];
--- a/packages/babel-plugin-minify-infinity/src/index.js
+++ b/packages/babel-plugin-minify-infinity/src/index.js
@@ -27,11 +27,16 @@ module.exports = function({ types: t }) {
         if (path.parentPath.isMemberExpression()) {
           return;
         }
-
         const bindingIds = path.parentPath.getBindingIdentifierPaths();
 
         for (const id of Object.keys(bindingIds)) {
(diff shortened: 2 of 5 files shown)


packages/babel-plugin-minify-infinity/__tests__/fixtures/expressions/expected.js
```diff
@@ -0,0 +1,6 @@
+let x = [1 / 0, 1 / 0];
+let y = [{
+  a: Infinity
```
Cyp: Should expect `  a: 1 / 0`, I think.

packages/babel-plugin-minify-infinity/src/index.js
```diff
@@ -27,11 +27,16 @@ module.exports = function({ types: t }) {
         if (path.parentPath.isMemberExpression()) {
           return;
         }
-
         const bindingIds = path.parentPath.getBindingIdentifierPaths();
```
Cyp: Seems to work. If I'm reading it right, could
```js
        for (const id of Object.keys(bindingIds)) {
          if (
            id === "Infinity" &&
            bindingIds[id] === path &&
            // ObjectProperty is common for ObjectExpression and ObjectPattern and only
            // one of them is a Binding, the other is simply a reference
            !path.parentPath.parentPath.isObjectExpression()
          ) {
            return;
          }
        }
```
be simplified to this?
```js
        if (
          bindingIds["Infinity"] === path &&
            // ObjectProperty is common for ObjectExpression and ObjectPattern and only
            // one of them is a Binding, the other is simply a reference
            !path.parentPath.parentPath.isObjectExpression()
          ) {
          return;
        }
```

PR #842, opened 2018-05-14
fix(mangle): handle inferred names for functions

### LINKED ISSUE #829: Anonymous functions sometimes given conflicting names, breaking code.
#### Input Code

index.js:
```js
#!/usr/bin/node
'use strict';

const babel = require('babel-core');

console.log(babel.transform(`
	'use strict';
	(() => {
		const con = console;
		const obj = {
			a() {
				con.log('Hi.');
			}
		};
		obj.a.log = str => {
			con.log('OUCH, should never get here! Should be "' + str + '".');
		};
		obj.a();
	})();
`, {presets: ['minify', 'env']}).code);
```

package.json:
```
{
  "name": "babelbug",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "",
  "private": true,
  "dependencies": {
    "babel-core": "^6.26.3",
    "babel-preset-env": "^1.6.1",
    "babel-preset-minify": "^0.4.0"
  }
}
```

#### Actual Output

```js
'use strict';(function(){var a=console,b={a:function a(){a.log('Hi.')}};b.a.log=function(b){
a.log('OUCH, should never get here! Should be "'+b+'".')},b.a()})();
```
Prints 'OUCH, should never get here! Should be "Hi.".'.

#### Expected Output

```js
'use strict';(function(){var a=console,b={a:function(){a.log('Hi.')}};b.a.log=function(b){
a.log('OUCH, should never get here! Should be "'+b+'".')},b.a()})();
```

Prints 'Hi.'.

#### Details
Note the incorrect `function a()` instead of `function()`, leading to the `con` variable being clobbered.

Same thing if using babel-preset-es2015 with babel-preset-minify.

Can't make a direct link to reproduce online – not sure why. But following this link, clicking on 'Plugins', on 'Only official plugins', typing 'mangle' and selecting minify-mangle-names reproduces it:
https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=JAcgrgzgpgBBAuAnAlgY3iA3AKGACjwEoYBeAPhgG9dhUB7AOwRnodKoBs6BzALhgAepCgEMOURPDwDCAXxzBajZnQBGAK3bVFwEUSo1FrAHRdueEAAlkxkIQWLZNeTTXrjI0z3YJEwgzomZhYA8gCqAMKWADRwABZ0YBwAJjAMUABuEjDcUPAwcRJQAIQwAMoJSamqsABEIDAA1HBITTAgtbb2zg5uHkQKsoQD2EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=es2015&prettier=false&targets=&version=6.26.0&envVersion=

Not sure whether this is a minify bug or a babel bug. Can't reproduce if running 'env' on the output of 'minify' or when running 'minify' on the output of 'env', only if doing both in the same step.

boopathi:
The t.NOT_LOCAL_BINDING symbol used in plugin-function-name
(https://github.com/babel/babel/blob/bd98041321c60737ddcacd1ad7e056ef6de31879/packages/babel-helper-function-name/src/index.js#L206)
is causing this issue.

This was added in babel/babel#3298

The problem is the binding identifier is NOT registered (skipped) during
crawl and the mangler finds the first safe identifier which ultimately
collides with this name during runtime.

So, we register the binding and don't respect the t.NOT_LOCAL_BINDING.

Fix #829

### DIFF
--- a/packages/babel-plugin-minify-mangle-names/src/index.js
+++ b/packages/babel-plugin-minify-mangle-names/src/index.js
@@ -149,8 +149,6 @@ module.exports = babel => {
          * - https://github.com/babel/minify/issues/829
          */
         BindingIdentifier(path) {
-          const { scope } = path;
-
           if (
             // the parent has this id as the name
             (path.parentPath.isFunctionExpression({ id: path.node }) ||
--- a/packages/babel-plugin-minify-mangle-names/src/index.js
+++ b/packages/babel-plugin-minify-mangle-names/src/index.js
@@ -140,6 +140,29 @@ module.exports = babel => {
         },
 
         /**
+         * This is required because after function name transformation
+         * plugin (part of es2015), the function name is NOT added to the
+         * scope's bindings. So to fix this issue, we simply add a hack to
+         * handle that case - fix it to the scope tree.
+         *
+         * Related:
+         * - https://github.com/babel/minify/issues/829
+         */
+        BindingIdentifier(path) {
+          const { scope } = path;
+
+          if (
+            // the parent has this id as the name
+            (path.parentPath.isFunctionExpression({ id: path.node }) ||
+              path.parentPath.isClassExpression({ id: path.node })) &&
+            // and the id isn't yet added to the scope
+            !hop.call(path.parentPath.scope.bindings, path.node.name)
+          ) {
+            path.parentPath.scope.registerBinding("local", path.parentPath);
+          }
+        },
+
+        /**
          * This is necessary because, in Babel, the scope.references
          * does NOT contain the references in that scope. Only the program
          * scope (top most level) contains all the references.
--- a/packages/babel-preset-minify/__tests__/minify-env-tests.js
+++ b/packages/babel-preset-minify/__tests__/minify-env-tests.js
@@ -237,7 +237,7 @@ describe("preset along with env", () => {
       }
     `,
     `
-      function _classCallCheck(a, b) { if (!(a instanceof b)) throw new TypeError(\"Cannot call a class as a function\"); }
+      function _classCallCheck(a, b) { if (!(a instanceof b)) throw new TypeError("Cannot call a class as a function"); }
 
       function bar() {
         var c = console;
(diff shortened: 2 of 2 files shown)


packages/babel-preset-minify/__tests__/minify-env-tests.js
```diff
@@ -196,4 +196,59 @@ describe("preset along with env", () => {
...
+            con.log(bar);
+          }
+        };
+      }
+    `,
+    `
+      function foo() {
+        var b = console;
+        return {
+          a: function d(c) {
```
j-f1: Shouldn’t this be `function a` so that its `.name` is correct?

packages/babel-plugin-minify-mangle-names/src/index.js
```diff
@@ -139,6 +139,29 @@ module.exports = babel => {
...
+         * This is required because after function name transformation
+         * plugin (part of es2015), the function name is NOT added to the
+         * scope's bindings. So to fix this issue, we simply add a hack to
+         * handle that case - fix it to the scope tree.
+         *
+         * Related:
+         * - https://github.com/babel/minify/issues/829
+         */
+        BindingIdentifier(path) {
+          const { scope } = path;
```
vigneshshanmugam: unused.

packages/babel-preset-minify/__tests__/minify-env-tests.js
vigneshshanmugam: Handled by `keepFnName` option.

boopathi: yes. Correct. Deadcode should have taken care of that (have to debug why it didn't in this case), and we come back to plugin ordering problem - https://github.com/babel/minify/issues/422

boopathi: I guess it's the same issue as this one - we have to add the same thing in DCE (register this binding before dce runs).

PR #848, opened 2018-05-15
fix(evaluate): handle when parentpath is removed

vigneshshanmugam:
+ Handled only for let/constant bindings, extending it to var binding as well.


chharvey:
good job in fixing this. however it looks like the fix has not been pushed to production. when you run `npm install` you get a lib file without the fix:
https://unpkg.com/babel-helper-evaluate-path@0.4.3/lib/index.js

please update the distributed/built package and publish to v0.4.4 thank you

keithws:
+1 need this in production.

PR #859, opened 2018-05-17
perf(builtins): run builtins transform on multi pass

vigneshshanmugam:
+ This improves the performance approximately from ~600ms to ~250ms on a 2mb file.


vigneshshanmugam:
I love where we have finally reached with the current set of optimisations. The average time went from ~600ms to ~80ms on a 2mb test file.

vigneshshanmugam:
For Lodash, we went from approx 100ms to 12ms. 

But out of it, 60-70ms was from path traversal context creation which is fixed through multi pass.

Issue #792, opened 2018-02-07
TypeError: Cannot read property 'add' of undefined


adamreisnz:
#### Input Code

```js

/**
 * Dependencies
 */
import $ from 'jquery';
import $doc from '../../shared/$doc.js';

/**
 * User panel service
 */
class UserPanelService {

  /**
   * Constructor
   */
  constructor() {

    //ID
    this.id = 'UserPanel';

    //Classes
    this.cToggle = 'UserPanelToggle';
  }

  /**************************************************************************
   * Initialisation
   ***/

  /**
   * Initialisation
   */
  init() {
    this.findElements();
    this.bindToggleHandlers();
    this.bindGlobalCloseHandler();
  }

  /**
   * Find elements
   */
  findElements() {
    this.$userPanel = $(`#${this.id}`);
  }

  /**
   * Bind toggle handlers
   */
  bindToggleHandlers() {
    $doc
      .on('click', `.${this.cToggle}`, () => {
        this.toggle();
      });
  }

  /**
   * Bind global close handler
   */
  bindGlobalCloseHandler() {
    $doc
      .on('click', event => {

        //Not visible?
        if (!this.isVisible()) {
          return;
        }

        //Get target
        const $target = $(event.target);
        const $parents = $target.parents(`.${this.cToggle}`);

        //Clicked on user panel
        if (this.$userPanel[0].contains($target[0])) {
          return;
        }

        //Clicked on toggle
        if ($target.hasClass(this.cToggle) || $parents.length > 0) {
          return;
        }

        //Hide
        this.hide();
      });
  }

  /**************************************************************************
   * Panel visibility
   ***/

  /**
   * Check if visible
   */
  isVisible() {
    return this.$userPanel.is(':visible');
  }

  /**
   * Toggle the user panel
   */
  toggle() {
    this.$userPanel.toggle();
  }

  /**
   * Hide the user panel
   */
  hide() {
    this.$userPanel.hide();
  }

  /**
   * Show the user panel
   */
  show() {
    this.$userPanel.show();
  }
}

/**
 * Export service singleton
 */
export default new UserPanelService();
```

Forgive me, it's using jQuery, I had to for this project 😩 

#### Actual Output
```
TypeError: app/components/app/header/user-panel.service.js: Cannot read property 'add' of undefined
    at ScopeTracker.addReference (/node_modules/babel-plugin-minify-mangle-names/lib/scope-tracker.js:55:36)
[...]

Issue #556, opened 2017-05-24
[BUG] Cannot read property 'add' of undefined


louisgv:
Here's my `.babelrc`:
<details>
 <summary>babelrc</summary>

```
{
  "presets": [
    "es2015", "stage-2", "stage-0", "react", "es2015-ie"
  ],
  "plugins": [
    "transform-runtime", "transform-decorators-legacy", "transform-class-properties"
  ],
  "compact": true,
  "env": {
    "test": {
      "plugins": ["istanbul"]
    },
    "production": {
      "presets": [
         "babili"
      ]
    }
  }
}
```
</details>

Running babel resulted in this error:

<details>

```
ERROR in ./src/components/Main/index.js
Module build failed: TypeError: /src/components/Main/index.js: Cannot read property 'add' of undefined
    at ScopeTracker.addReference (/node_modules/babel-plugin-minify-mangle-names/lib/scope-tracker.js:55:36)
    at ReferencedIdentifier (/node_modules/babel-plugin-minify-mangle-names/lib/index.js:190:28)
    at newFn (/node_modules/babel-traverse/lib/visitors.js:318:17)
    at bfsTraverse (/node_modules/babel-plugin-minify-mangle-names/lib/bfs-traverse.js:34:43)
    at Mangler.collect (/node_modules/babel-plugin-minify-mangle-names/lib/index.js:269:9)
    at Mangler.run (/node_modules/babel-plugin-minify-mangle-names/lib/index.js:70:14)
    at PluginPass.exit (/node_modules/babel-plugin-minify-mangle-names/lib/index.js:589:19)
    at newFn (/node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (/node_modules/babel-traverse/lib/path/context.js:117:8)
    at TraversalContext.visitQueue (/node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitSingle (/node_modules/babel-traverse/lib/context.js:108:19)
    at TraversalContext.visit (/node_modules/babel-traverse/lib/context.js:192:19)
    at Function.traverse.node (/node_modules/babel-traverse/lib/index.js:114:17)
    at traverse (/node_modules/babel-traverse/lib/index.js:79:12)
[...]

PR #860, opened 2018-05-20
Add missing scopes when adding reference to scope tracker

devongovett:
Fixes "Cannot read property 'add' of undefined" error. #792 #556


devongovett:
Ok I found the problem in my code. Needed to `cloneDeep` a node before inserting it multiple times.

Issue #862, opened 2018-05-22
Missing "file" property result source map (gulp-babel-minify)


egasato:
The pipe crashes when the "file" property is missing in the "result" variable.

**To Reproduce**
My gulpfile.js serves as an example.

```js
gulp.task("babel:server", function(_done) {
	return gulp.src("lib/*.mjs")
		.pipe(plugins.plumber(function(err) {
			log.error(err);
			this.emit("end");
		}))
		.pipe(plugins.sourcemaps.init())
		.pipe(plugins.babelMinify())
		.pipe(plugins.sourcemaps.write("."))
		.pipe(gulp.dest("build"));
});
```

**Actual Output**

I simply run `gulp babel:server` and this is what happens:

```
[14:34:36] Using gulpfile /run/media/esau/Shared/Universidad/PTI/Weether/iot-station/gulpfile.js
[14:34:36] Starting 'babel:server'...
/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/vinyl-sourcemaps-apply/index.js:37
    throw e;
    ^

Error: Source map to be applied is missing the "file" property
    at assertProperty (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/vinyl-sourcemaps-apply/index.js:36:13)
    at applySourceMap (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/vinyl-sourcemaps-apply/index.js:15:3)
    at DestroyableTransform._transform (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/gulp-babel-minify/lib/index.js:75:9)
    at DestroyableTransform.Transform._read (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/readable-stream/lib/_stream_transform.js:184:10)
    at DestroyableTransform.Transform._write (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/readable-stream/lib/_stream_transform.js:172:83)
    at doWrite (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/readable-stream/lib/_stream_writable.js:428:64)
    at writeOrBuffer (/run/media/esau/Shared/Universidad/PTI/Weether/iot-station/node_modules/readable-stream/lib/_stream_writable.js:417:5)
[...]

PR #863, opened 2018-05-22
Fix issue #862

egasato:
The following changes were applied:

  - Added a condition where the property "file" is checked. If it
    doesn't exist it is copied from the original source file.

P.S: I just saw that the indentation of the closing brace is wrong (Vim's fault).


egasato:
I'm sorry but I'm new to the Pull Request world. Is there a way to update the code so that Travis CI doesn't fail? The log says it's because of the indentation I previously mentioned.

boopathi:
Thanks

PACelah:
Sweet

PR #891, opened 2018-08-03
fix: missing dependency babel-helper-evaluate-path and add eslint plugin

### LINKED ISSUE #884: Some packages have missing dependencies
**Describe the bug**
* [babel-plugin-minify-guarded-expressions](https://github.com/babel/minify/blob/9939bca95c66021e05528bb3b0eada0bc595d4c5/packages/babel-plugin-minify-guarded-expressions/src/index.js#L2) is missing dependency `babel-helper-evaluate-path`
* [babel-plugin-minify-simplify](https://github.com/babel/minify/blob/9939bca95c66021e05528bb3b0eada0bc595d4c5/packages/babel-plugin-minify-simplify/src/helpers.js#L4) is missing dependency `babel-helper-evaluate-path`

These were the ones I ran into but there might be more.

christophehurpeau:
fixes #884

Unfortunately, the rule "import/no-extraneous-dependencies" doesn't work if eslint is run from the root.
I changed the scripts `lint` and `fix` to run eslint in root then eslint in each packages with lerna, add added `/packages` in the eslintignore root. Is that acceptable ?


boopathi:
I'm not sure this is worth running eslint per package. If the eslint-plugin-import does not support it, I wouldn't use it in this case. We don't have that many changing packages or new packages coming in. For now, I'd prefer manually adding it.

Issue #879, opened 2018-07-02
babel-plugin-minify-dead-code-elimination only useful for entry scripts


kingback:
**Describe the bug**

`babel-plugin-minify-dead-code-elimination` only useful for entry scripts, the dependencies is not working.

**To Reproduce**

`index.js`

```js
import { setTitle } from './utils';
const isWeb = true;

setTitle('test');

if (isWeb) {
  console.log(isWeb);
} else {
  console.log(!isWeb);
}
```

`utils.js`

```js
const isWeb = true;

export function setTitle(title) {
  if (isWeb) {
    document.title = title;
  } else {
    console.log(title);
  }
}
```

**Actual Output**

`index.js`, correct

```js
import { setTitle } from './utils';
const isWeb = true;
setTitle('test');
console.log(isWeb);
```

`utils.js`, **uncorrect**

```js
const isWeb = true;

export function setTitle(title) {
  if (isWeb) {
    document.title = title;
  } else {
    console.log(title);
  }
}
```

**Expected Output**

`utils.js`

```js
const isWeb = true;
export function setTitle(title) {
  document.title = title;
}
```

**Configuration**

How are you using babel-minify?

webpack.config.js

```js
const path = require('path');

module.exports = {
  mode: 'production',
  entry: './src/index/index.js',
  output: {
    path: path.resolve(__dirname, 'build'),
    filename: 'bundle.js'
  },
  optimization: {
    minimize: false
  },
  module: {
    rules: [
      {
        test: /\.(js|jsx)$/,
        loader: 'babel-loader',
        options: {
          presets: ['env'],
          plugins: ['minify-dead-code-elimination']
        }
      }
    ]
  }
};
```

package.json

```json
{
  "name": "test-dce",
  "version": "1.0.0",
  "description": "test dead code elimination",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "dev": "webpack-dev-server --mode development",
    "build": "webpack --mode production"
  },
  "repository": {},
  "keywords": [],
  "author": "huya.nzb@alibaba-inc.com",
  "license": "MIT",
  "devDependencies": {
    "babel-core": "^6.26.3",
    "babel-loader": "^7.1.4",
    "babel-plugin-minify-dead-code-elimination": "^0.4.3",
    "babel-preset-env": "^1.7.0",
    "webpack": "^4.14.0",
    "webpack-cli": "^3.0.8",
[...]

PR #892, opened 2018-08-03
fix: issue #879 dead code elimination with const in outer scope

christophehurpeau:


packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -80,6 +80,10 @@ function evaluateIdentifier(path) {
     }
   }
 
+  if (binding.kind === 'const') {
+    return path.evaluate();
```
boopathi: this will skip the TDZ check below for `const`

christophehurpeau: do you have an code example where this would be a problem ? The solution would be to add `binding.kind !== 'const' &&` in `if (shouldDeoptBasedOnScope(binding, path)) {` ?

christophehurpeau: (sorry about the outdated I force-pushed to fix the tests...)

christophehurpeau:
@boopathi please would you have some time to review ? thanks :)

vigneshshanmugam:
@boopathi Check once, Lets get this in

packages/babel-helper-evaluate-path/src/index.js
```diff
@@ -109,9 +109,14 @@ function evaluateIdentifier(path) {
 // if the binding is in program scope
 // all it's references inside a different function should be deopted
 function shouldDeoptBasedOnScope(binding, refPath) {
+  if (binding.kind === "const") {
+    return false;
+  }
+
```
boopathi: why only `const` ?. why does this repeat twice - one below?

christophehurpeau: because let variable values can change ?

PR #900, opened 2018-09-20
Upgrade to babel@7 and lerna@3

pksunkara:
Lerna 2 has license distribution issues which were fixed in Lerna 3. Also the same thing happened with Babel, which was fixed in Babel 7.

So, basically, once this PR is merged and you guys publish all the packages, the license issues will be fixed.


pksunkara:
@boopathi Could you please republish after this is merged? Thanks.

pksunkara:
The canary publish in master is broken. My bad. Will send a PR fixing it.

PR #901, opened 2018-09-21
Fix canary publishing

pksunkara:
Upgrading to lerna 3 broke the canary publishing. I missed correcting it in PR #900. So, here's the fix.


pksunkara:
@boopathi Hey, can I get this reviewed?

Issue #906, opened 2018-09-25
Lerna Publishing change


pksunkara:
@boopathi Now that 0.5.0 is released, should the lerna canary publish change to patch version? Or we still want to publish minor version for canary?

PR #907, opened 2018-09-26
chore: Don't do shallow clones on travis

pksunkara:
Lerna canary doesn't work correctly with shallow clones.

From https://github.com/lerna/lerna/issues/1583#issuecomment-424596666.

@boopathi I can also incorporate #906 if you decide on it.

From #900 

> So, basically, once this PR is merged and you guys publish all the packages, the license issues will be fixed.

You haven't published all the packages. Only the packages which got changed, got published. Which means, the licensing issues still exist for older packages. Can you do a `lerna publish --force-publish` to fix this? (A new patch version should be good enough). Very important for us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 git:
-  depth: 10
+  depth: false
   submodules: false
 sudo: false
 language: node_js
 cache: yarn
-
 node_js:
   - "10"
   - "8"


pksunkara:
Canary releases depend on finding the distance from the recent tag so that they can be correctly versioned according to semver.

.travis.yml
```diff
@@ -4,7 +4,8 @@ git:
 sudo: false
 language: node_js
 cache: yarn
-
+git:
+  depth: false
```
boopathi: there is `git depth:10` at line 1,2 . Change that instead of adding again.

pksunkara:
Fixed. Please don't forget to do a 0.5.1 release and publish all the packages. Some of them still don't have licenses in the distributed versions.

pksunkara:
@boopathi Some of the packages like `babel-plugin-minify-guarded-expressions` still doesn't have licenses since you haven't published them with 0.5.0. Please fix this.

ajkl2533:
@boopathi Hi could you please publish the new version of packages? This is unfortunately real blocker for legal use of some other packages, like Storybook. 
Thanks 👍

opichals:
@boopathi As @pksunkara suggested in order to fix the missing license files a `lerna publish --force-publish` is necessary to force a new patch version releases.

PR #917, opened 2018-10-18
remove subsequent nodes after 100% truthy condition

### LINKED ISSUE #898: Incorrect output in minify-dead-code-elimination
**Describe the bug**

A strange edge case code does not get eliminated correctly.

**To Reproduce**

Minimal code to reproduce the bug

```js
function b() {}

function a() {
  if ('production' === 'production') {
    return
  }

  var c = b()

  if (!c) {
    return
  }
}
```

**Actual Output**

If there is no Error thrown,

```js
function b() {}

function a() {

  var c = b();
}
```

**Expected Output**

```js
function b() {}

function a() {

}
```
**Configuration**

How are you using babel-minify?

`babel dependency`

babel version : `7.0.0-beta.3`

Used stand alone using the Node Api + babel-core

```json5
{
  plugins: [],
  presets: []
}
```

**Additional context**

Encountered this [here](https://github.com/atlassian/react-beautiful-dnd/blob/b276e4780c42ad662400631c9f315edc791a9cd8/src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx#L53-L77)

Tried to recreate it as minimal as possible using only babel-core and the dead-code plugin

msn0:
Fixes #898

### DIFF
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/issue-898/actual.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/issue-898/actual.js
@@ -1,0 +1,8 @@
+function a() {
+  if (true) {
+    return;
+  }
+
+  var c = b();
+  return c;
+}
--- a/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/issue-898/expected.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/issue-898/expected.js
@@ -1,0 +1,1 @@
+function a() {}
--- a/packages/babel-plugin-minify-dead-code-elimination/src/index.js
+++ b/packages/babel-plugin-minify-dead-code-elimination/src/index.js
@@ -879,12 +879,20 @@ module.exports = ({ types: t, traverse }) => {
           }
 
           // we can check if a test will be truthy 100% and if so then we can inline
-          // the consequent and completely ignore the alternate
+          // the consequent, completely ignore the alternate and remove subsequent nodes
           //
           //   if (true) { foo; } -> { foo; }
           //   if ("foo") { foo; } -> { foo; }
           //
           if (evalResult.confident && evalResult.value) {
+            if (path.parentPath.isBlockStatement()) {
+              let i = 1;
+              while (path.key + i < path.container.length) {
+                path.getSibling(path.key + 1).remove();
+                i++;
+              }
+            }
+
             path.replaceWithMultiple([
               ...replacements,
               ...toStatements(consequent),


msn0:
What a fail 😄 
I'll fix it.

Cyp:
That seems better. It would be good if it could also handle this case somehow:
```js
var v = 43;

function s() {
  s.func = func;
  if (true) {
    return;
  }
  console.log('no-hi');

  var v = (console.log('no-hi2'), 42);
  function func() {
    v = (v | 0) + 1;
    console.log('v is ' + v);
  }
  console.log('test2');
}

s();
s.func();
```
What happens:
```js
"use strict";var v=43;function s(){s.func=func;function func(){v=(v|0)+1;console.log("v is "+v)}}s();s.func();
// prints v is 44
```
What should happen:
```js
"use strict";var v=43;function s(){s.func=func;var v;function func(){v=(v|0)+1;console.log("v is "+v)}}s();s.func();
// prints v is 1
```
Should probably work with `{var v}` as well as just `var v`.

msn0:
ok, looks like all declarations should be hoisted somehow. I need to figure out how to do it ;)

Cyp:
It seems to work now, although some var declarations can be a bit redundant:
```js
var g;
function a() {
  var h = ' ';
  g += h + i + j + k + l + m + n;
  if (true) {
    return;
    var i = 1;
  } else {
    var j = 2;
  }
  var k = 3;
  var l = 4;
  while (true) {
    var m = 5;
    if (false) {
      var n = 6;
    }
  }
}
```
```js
"use strict";var g;function a(){var h=" ";g+=h+i+j+k+l+m+n;var i=1;var j;var h,i,j,k,l,m,n;
var h,i,j,h,i,j,k,l,m,n,l,m,n}
```
The function `a()` does declare all variables at least once and correctly set `g = 'undefined undefinedundefinedundefinedundefinedundefinedundefined'`, so it's not wrong.

Swapping the `if (true) A; else B;` to `if (false) B; else A;`:
```js
var g;
function a() {
  var h = ' ';
  g += h + i + j + k + l + m + n;
  if (false) {
    var j = 2;
  } else {
    return;
    var i = 1;
  }
  var k = 3;
  var l = 4;
  while (true) {
    var m = 5;
    if (false) {
      var n = 6;
    }
  }
}
```
```js
"use strict";var g;function a(){g+=" "+void 0+void 0+void 0+void 0+m+n}
```
It inlines most of the vars, but misses m and n.

msn0:
With your last example it looks bad imho. `var i` should not be initialised with `1` and these redundant variable declarations should not happen.

msn0:
@Cyp I believe I've covered your recent examples ;) Could you please take a look?

msn0:
@Cyp `a1` seems to be fixed right now. `a2` and `a3` looks similar nevertheless it's not related to ifStatement. Anyway I'll try to look on it - maybe there's some simple fix for these.

msn0:
I'll create a separate issue(s) for `a2` and `a3` cases. I think it's unrelated to IfStatement. I've got a fix already for `a2`, rather simple.

Cyp:
Not finding any remaining issues.

Not an issue (and nothing to do with the patch), but this isn't fully simplified:
```js
function d1(x) {
  if (x) ; else ;
  if (x) {} else ;
  if (x) ; else {}
  if (x) {} else {}
}
```
```js
"use strict";function d1(x){if(x);else;if(x){}else;if(x);if(x){}}
```

msn0:
@boopathi please take a look. Should I change something or abandon it?

coreyfarrell:
What about `throw`?  Examples:
```js
function f1() {
  if (true) {
   throw new Error('throw');
  }

  return 'unreachable';
}

function f2() {
  if (true) {
    throw new Error('throw');
  } else {
    return 'return';
  }

  return 'unreachable';
}
```

PR #918, opened 2018-10-22
chore(minify): remediate CVE-2018-3721: update lodash to latest, remove use of deprecated module packages

Stephanemw:
What:
Updated use of lodash to lodash/somefunc instead of lodash.somefunc:
* upgraded lodash deps to ^4.17.11
* updated requires from lodash.somefunc to lodash/somefunc
* updates snapshots

re-ran bootstrap, build and test

Why:
Updated to remediate CVE-2018-3721 - where modularized packages are specifically vulnerable, and as modularized packages are deprecated, no updates have been released in about 2 years.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ coverage
 lib/
 *.log
 *~
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ coverage
 lib/
 *.log
 *~
+.idea/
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ coverage
 lib/
 *.log
 *~
+.idea/
```
boopathi: this is not required. you can add it to your global gitignore

Stephanemw: thanks for the feedback, correcting in next commit

Stephanemw:
@boopathi thanks for merging! Would you know when this is expected to be included in a non-alpha release?

PR #953, opened 2019-05-27
Dead code elimination removes unused import bindings

### LINKED ISSUE #586: Remove unused imports
```js
// Map won't be eliminated when bundling
import Map from 'lodash.map'
```
There are non-pure modules with side effects. Is there any way to remove the unused import regardless of side-effect.
[Possible related rollup feature](https://github.com/rollup/rollup/pull/1388)
Thank you.

conartist6:
Fixes #586


packages/babel-plugin-minify-dead-code-elimination/__tests__/fixtures/unused-import-bindings/expected.js
```diff
@@ -0,0 +1,6 @@
+import "file";
+import "test-variable-names";
+import "weird-test-variable-names";
+import bark from "dogs";
+import { woof } from "dogs";
```
j-f1: Should these two be combined?

```js
import bark, { woof } from "dogs";
```

conartist6: Yes, but that's trickier. If we want to support it I'd rather do it was a separate PR.

conartist6:
@boopathi hello?

PR #963, opened 2019-08-15
Add missing `babel-helper-evaluate-path` dependency

feross:
At the moment, the `babel-plugin-minify-guarded-expressions` package is calling `require('babel-helper-evaluate-path')` without actually depending on this package. It only works right now because of package hoisting.

An exception is thrown when npm does not flatten the dependency tree. You can trigger this to verify it by forcing npm to install packages without flattening (the npm@2 nested `node_modules` approach) by running: `npm install --legacy-bundling`


feross:
Note: There might be other instances of this. This is the first one I found that crashed `babel-minify`.

hzoo:
Thanks for the ping online @feross. Ah looks like there was a previous PR for this, so goin to merge that one! https://github.com/babel/minify/pull/957/

feross:
Great, thanks!

PR #970, opened 2019-11-12
fix: an error occurred if includes environment variable assignment expression

mochiya98:
The following code reproduce the bug:

```js
if(process.env.NODE_ENV !== "production"){
  process.env.NODE_ENV = "development"; // an error occurred
  // ...
}
```
```bash
babel --plugins transform-inline-environment-variables script.js
```
> TypeError: Property left of AssignmentExpression expected node to be of a type ["LVal"] but instead got "StringLiteral"

### DIFF
--- a/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
+++ b/packages/babel-plugin-transform-inline-environment-variables/__tests__/inline-env-var-test.js
@@ -24,6 +24,16 @@ describe("inline-env-plugin", () => {
   );
 
   thePlugin(
+    "should not inline environment variables if it is on left side of assigment expression",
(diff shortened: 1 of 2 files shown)


koenoe:
Will this ever be merged?

tockn:
+1

clarle:
I tested this both independently and with a few packages that rely on Minify, and all seems to be fine.  I'll merge this in to get this into canary so more people will have an opportunity to test it further and that it doesn't break any unexpected behavior.

mochiya98:
Thanks!

hzoo:
^ Just for some context, @clarle reached out and I'm giving contributor access to help maintain the repo, so thanks!

PR #1002, opened 2021-03-22
Adding the required option

hecflores:
Adding the **required** property. This property will be used to define if environment variables are required and also to choose which ones are required.


hecflores:
@boopathi  let me know if there is anything needed from me for this.

hecflores:
@boopathi is there a reason my build is failing... Is there some other PR that is working on the dependency issue?

PR #1032, opened 2022-05-04
Do not inline `this`

JLHwung:
From now on just bail on `ThisExpression`.

For further improvement ideas, see https://github.com/babel/minify/issues/1028#issuecomment-1114982399

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
     "watch": "gulp watch"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


nicolo-ribaudo (approved):
I think I have publish access for these packages, I'll figure out how to do it.

aweary:
@nicolo-ribaudo I don't think this fix ever got published?

PR #1047, opened 2023-02-26
[1046] Maintain source mapping for merged sibling vars.

seanmorris:
The plugin babel-plugin-transform-merge-sibling-variables does not maintain source mapping when merging sibling variable declarators into a single declaration after babel-plugin-transform-modules-commonjs has been applied.

To Reproduce:
Minify a file that has been processed with babel-plugin-transform-modules-commonjs. If the original file has more than one import statement, the mapping will be lost.

Minimal code to reproduce the bug: https://github.com/babel/minify/issues/1046


seanmorris:
@boopathi @nicolo-ribaudo hey are there any updates on this?

